PR #184, opened 2015-02-02
ALTAIR: Fix initialization of uninstalled memory

rscottbailey:
Reads of uninstalled memory are supposed to return 0377 (0xff). However,
memory bytes beyond the installed limit were being initialized with 0
instead of 0377. Fix comment and the actual initialization loop.

### DIFF
--- a/ALTAIR/altair_cpu.c
+++ b/ALTAIR/altair_cpu.c
@@ -25,6 +25,7 @@
 
    cpu          8080 CPU
 
+   02-Feb-15    RSB     Fixed initialization of "uninstalled" memory
    19-Mar-12    RMS     Fixed data type for breakpoint variables
    08-Oct-02    RMS     Tied off spurious compiler warnings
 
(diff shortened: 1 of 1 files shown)


markpizz:
Hi Scott,

Thanks for this.

Can you redo it conforming to the following guidelines:

1) Pull https://github.com/simh/simh master just prior to your commit.
2) Prefix your commit message with the component affected.  i.e. "ALTAIR: Fix initialization of uninstalled memory"
3) In general, we no longer change comments at the top of source file since the commit history contains the best information AND (although not relevant in this case) when changes are being made to the same files by different folks, changes to the comments at the top of source files cause merge conflicts which need to be handled manually.

Cancel this pull request and generate a new one immediately after generating your commit.  No problem pushing it to the master branch in your github repo.  To achieve this you may need to delete your github.com/rscottbailey/simh repo and fork it again.

I haven't been 100% consistent with enforcing these guidelines, but if I don't give you the feedback it will never happen.

Thanks.

- Mark

markpizz:
Fixed with cherry-pick.  Forced to learn a new workflow.

Thanks.

PR #188, opened 2015-02-13
BESM6: Enhanced the front panel, code cleanup.

leobru:
No warnings in BESM6/*.c with -Wall

### DIFF
--- a/BESM6/besm6_cpu.c
+++ b/BESM6/besm6_cpu.c
@@ -559,11 +559,7 @@ static void cmd_033 ()
     switch (Aex & 04177) {
     case 0:
         /*
-         * Using an I/O control instruction with Aex == 0
-         * after issuing a 033 instruction with a non-zero Aex
-         * to send data to a device was required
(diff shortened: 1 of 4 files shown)


markpizz:
Thanks for the update.

It seems that a tab character has leaked into your recent change in besm6_printer.c.  

I've recently noticed that the libSDL source has the last line of source files that contain:
/* vi: set ts=4 sw=4 expandtab: */
Do you know if this is something that vi (vim) dynamically recognizes or is it just advice to someone thinking about editing things???

Rhialto:
On Fri 13 Feb 2015 at 05:49:35 -0800, Mark Pizzolato wrote:
> /* vi: set ts=4 sw=4 expandtab: */
> Do you know if this is something that vi (vim) dynamically recognizes
> or is it just advice to someone thinking about editing things???

Yes, if the vi(m) setting "modeline" is on, it will perform those
settings automatically. Exactly how it recognises it (what the precise
format has to be) has varied a bit over time, becoming more strict.

PR #190, opened 2015-02-17
SDS: Allow interrupts to report I/O errors

markemmer:
As coded, the simulator would stop and report an I/O error, such as reading past EOF on a paper-tape disk file. Instead, it is proper to report the error via the interrupt logic for further processing by the operating system.


markpizz:
Hi Mark,

This change has a few issues:
1) You didn't pull before committing.
2) MORE IMPORTANTLY, if you want the simulator (CPU and devices) to handle issues of any particular category internally (with interrupts or device state), then the simulator components should do all of these things AND then return SCPE_OK.  SCPE_IOERR is for an unexpected issue that occurred within the simulator.  The main event dispatch loop in sim_instr() should exit or on any status value other than SCPE_OK.

- Mark

markemmer:
Will close this request and try something different.

PR #201, opened 2015-03-31
SCP: future-proof save/restore logic

rscottbailey:
Let me know if I'm getting better at this. ;-)

This is a small proactive change suggested to head off future programming errors in the save/restore routines.


rscottbailey:
I see I didn't get this right -- attempting to withdraw it for a retry. :-p

PR #209, opened 2015-04-21
Add a makefile to cross-compile simh to Windows on POSIX systems

Low-power:
Using the separate mingw.mk to cross-compile by typing ```make -f mingw.mk [<options>]```
The default compiler name is ```i586-mingw32msvc-gcc```; to change it by using option ```GCC=<...>```
Separated the source file list and the build rules to 2 new files sources.mk and rules.mk from original makefile


markpizz:
Compiling under MinGW has been documented for years.  One merely invokes make with:

     C:\Users\Joe> build_MinGW {make arguments}

{make arguments} can be any combination of target names, GCC=, DEBUG=, etc.

What problem does this solve?

Low-power:
Sorry I'm continue modify the code but forgot to create a new branch; I created a new pull request for this makefile change: https://github.com/simh/simh/pull/210

PR #210, opened 2015-04-22
Add a makefile to cross-compile simh to Windows on POSIX systems

Low-power:
Using the separate mingw.mk to cross-compile by typing ```make -f mingw.mk [<options>]```
The default compiler name is ```i586-mingw32msvc-gcc```; to change it by using option ```GCC=<...>```
Separated the source file list and the build rules to 2 new files sources.mk and rules.mk from original makefile


Low-power:
The build_mingw.bat works under on Windows only, but the MinGW toolchain is available on many non-Windows platforms for producing Windows executable. So this makefile could make simh for Windows on GNU/Linux for example.

markpizz:
The ultimate goal of your efforts are to provide ANOTHER way for users to get simulator binaries that can run on Windows.

Currently Windows users can choose from one of 3 different ways to get simulator binaries:

1) Build on their systems using the MinGW compiler environment
2) Build on their systems using any version of Visual Studio starting with Visual Studio 2008
3) Download pre-built binaries which are provided here on github for all simh versions and on simh.trailing-edge.com for older versions.

Adding another way to do this doesn't solve any problem that isn't already solved while at the same time introduces any number of potential complexities to a test matrix attempting to support this going into the future.

PR #211, opened 2015-04-25
AltairZ80: Fixed keyboard polling on non UNIX platform and daylight savi...

psco:
...ngs time issue with CP/M 3 clock


markpizz:
Hmm..   I can't see how the changes to any of the files in this commit (except altairz80_sio.c) really change anything.  As long as sim_poll_kbd gets called, stop_cpu will be set and the change to stop_cpu will cause sim_process_event() to return SCPE_STOP immediately as the reason.

The change to altairz80_sio.c suggests that sim_process_event() should check stop_cpu before exiting with SCPE_OK and if it is set return SCPE_STOP.  This change (to sim_process_event()) would make the change to altairz80_sio.c be unnecessary as well.

psco:
You are right, there is no functional change (I had some issue with control-E yesterday, but now they are gone, even with the previous version) - I will submit a cleaned up version shortly.

PR #212, opened 2015-04-25
AltairZ80: Simplified polling for stop CPU on non UNIX platform

psco:


markpizz:
Polling for ^E on each of these simulated systems EACH time any event for any device is processed is quite expensive.  Other simulators generally do this at most once per clock tick, often once every few clock ticks.

psco:
Actually this is the current behavior (only relevant to Windows) but I will check how the polling can be reduced.

markpizz:
I understand that Windows is the only affected platform. I was drawn to observe this due to the existence of the #ifndef UNIX_PLATFORM bracketing.  The goal is that simulators don't contain platform specific ifdefs.  Avoiding it on non-windows avoids a lot of useless overhead which won't ever produce any result.  A better solution would be to only poll relatively infrequently and thus there wouldn't be measurable overhead from excessive polling and ifdefs wouldn't be necessary.

PR #221, opened 2015-06-08
Added section for centos

abruzzim:
The uml utilties package is part of EPEL and named uml_utilities and not uml-utilities.


markpizz:
Merged via rebase

PR #230, opened 2015-08-08
Update s100_mdsad.c

deramp5113:
The source file to be updated provides emulation of the North Star DD floppy disk controller. The proposed change steps the head only when the step bit in the order byte transitions from 1 to 0. This is the way the actual hardware functions. Previously, the emulation would step any time an order was issued and the step bit was 1. This behavior caused some original software (e.g., Lifeboat CP/M 2.2) to end up on an incorrect track.

### DIFF
--- a/AltairZ80/s100_mdsad.c
+++ b/AltairZ80/s100_mdsad.c
@@ -101,6 +101,7 @@ typedef struct {
     uint8 ss;       /* Specifies the side of a double-sided diskette. The bottom side (and only side of a single-sided diskette) is selected when SS=0. The second (top) side is selected when SS=1. */
     uint8 dp;       /* has shared use. During stepping operations, DP=O specifies a step out and DP=1 specifies a step in. During write operations, write procompensation is invoked if and only if DP=1. */
     uint8 st;       /* controls the level of the head step signal to the disk drives. */
+    uint8 pst;      /* value of step signal (st) on previous order */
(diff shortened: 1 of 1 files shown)


markpizz:
Hi there,

Can you please put your full name on your Github profile and in your git configuration so your name appears on the commit?  The above explanation would be perfect comments as part of the commit message.  The commit message should have a first line which starts with "AltairZ80: Fix head stepping to only occur on 1 to 0 transitions" with an intervening blank line and the richer explanation and justification mentioned above..  

Thanks.

- Mark

deramp5113:
Mark,
Thanks for getting back to me in response to the pull request. I've submitted numerous changes to AltairZ80 via Peter Schorn and Howard Harte over the last couple of years - primarily fixing and/or adding disk emulations. Both of them agreed it was time for me to figure out what the hell Github is :)
Mike
 


     On Saturday, August 8, 2015 3:14 PM, Mark Pizzolato <notifications@github.com> wrote:
   

 Hi there,Can you please put your full name on your Github profile and in your git configuration so your name appears on the commit? The above explanation would be perfect comments as part of the commit message. The commit message should have a first line which starts with "AltairZ80: Fix head stepping to only occur on 1 to 0 transitions" with an intervening blank line and the richer explanation and justification mentioned above.. Thanks.   
   - Mark

PR #231, opened 2015-08-08
AltairZ80: Fix head stepping to only occur on 1 to 0 transitions.

deramp5113:
The source file to be updated provides emulation of the North Star DD
floppy disk controller. The change steps the head only when the step bit
in the order byte transitions from 1 to 0. This is the way the actual
hardware functions. Previously, the emulation would step any time an
order was issued and the step bit was 1. This behavior caused some
original software (e.g., Lifeboat CP/M 2.2) to end up on an incorrect
track.

### DIFF
--- a/AltairZ80/s100_mdsad.c
+++ b/AltairZ80/s100_mdsad.c
@@ -101,6 +101,7 @@ typedef struct {
     uint8 ss;       /* Specifies the side of a double-sided diskette. The bottom side (and only side of a single-sided diskette) is selected when SS=0. The second (top) side is selected when SS=1. */
     uint8 dp;       /* has shared use. During stepping operations, DP=O specifies a step out and DP=1 specifies a step in. During write operations, write procompensation is invoked if and only if DP=1. */
     uint8 st;       /* controls the level of the head step signal to the disk drives. */
+    uint8 pst;      /* value of step signal (st) on previous order */
     uint8 ds;       /* is the drive select field, encoded as follows: */
                     /* 0=no drive selected
                      * 1=drive 1 selected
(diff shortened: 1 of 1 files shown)


markpizz:
Hi Mike,

You're making progress and are very close at this time.  Your account looks good and the commit (by it self) looks good.

In general, I really prefer to simply pull minor updates like this.  I'll usually do this directly IF the repo I'm pulling from (yours in this case) is fully up to date.  The problem here is that the github/simh/simh repo (more specifically the master branch) has been changed since you cloned it so the pull operation in this state can't merely be a pull of your commit since a merge commit will also be necessary.  

In general, this is avoided if you (as the external party making the pull request) does a pull of the github/simh/simh  master branch before you commit your change and then you immediately generate the pull request.  Of course this process isn't 100% interlocked and a new commit could come in before you make your commit and generate the pull request, but it works almost all the time given the rather low commit rate on this repo.

Does this make sense?  Do you know what to do to fix it?  If not, please contact me directly at mark@infocomm.com and I'll walk you through the steps.

- Mark

deramp5113:
Good - glad I'm getting somewhere! Haven't worked under a source code control system since the late 80's. 

I cloned, committed, and did the pull request within a 15-30 minute period. Surprised I was out of sync. The actual source file was from a clone this morning, but I wouldn't think the source files have info about the master branch state do they?

Do you want me to redo it? No big deal if it would be easier for you.

Mike

markpizz:
Well, if it really was a 15-30 minute window, I just happened to have accepted a pull request from Peter right after you cloned.

In any case, to fix, 1) delete your github simh repo, 2) save the revised file someplace, 3) Fork the github.com/simh/simh repo to your account again and clone the result to your local system. 4) put the revised file back into your working directory, 5) generate the same commit you most recently did and 6) generate a new pull request.

PR #256, opened 2015-12-14
Avoiding conditional directives that break statements

RomeroMalaquias:


markpizz:
Hmmm...  

How did you happen to encounter this as a problem?

RomeroMalaquias:
Hi @markpizz, i'm doing a research about the discipline of conditional directives, and i've passed this repository on a tool that find undisciplined annotations.
I'm trying to know if developers care about this kind of refactoring.

markpizz:
Interesting.

By conditional directives you mean #if, #else, #endif conditionals, right?

The whole conditional expression that is there is somewhat complicated and already a little hard to read and reading it only makes sense to someone who is intimately familiar with the details of many of the things going on in the timing support code.  The proposed change you're making doesn't make it any easier to read or understand what is going on, and that is what I care about.  The async clock code which is in this module doesn't completely work yet and I'm working on a local branch to finish it.

I'm a little surprised that this was the only expression in this repository that was flagged by your tool.

RomeroMalaquias:
Yep, it's about #if, #else, #endif.
The idea is always to encompass only complete/entire statements/expressions with preprocessor conditional directives. It is something that is getting popular in many different projects (started mainly on the Linux Kernel).

We agree totally that breaking statements with directives influences code quality, i.e., understanding, maintainability and error proneness. We found some evidence based on interviews with developers and an online survey regarding this negative effect on quality.

As you can see on this [paper](http://www.cs.cmu.edu/~ckaestne/pdf/ecoop15.pdf).

There are more expressions, but i took this to try a refactoring and see you what you think about this.

Don't you agree with that?

markpizz:
I don't like the way your suggested change was written since, although it produces the correct result, it makes the details harder for me to understand.  If I were to rewrite that code block it would look like:

    t_bool sim_idle (uint32 tmr, t_bool sin_cyc)
    {
    static uint32 cyc_ms = 0;
    uint32 w_ms, w_idle, act_ms;
    int32 act_cyc;
    t_bool test_cant_idle;
    
    test_cant_idle = (!sim_idle_enab);                      /* idle disabled? */
    #if defined(SIM_ASYNCH_IO) && defined(SIM_ASYNCH_CLOCKS)
    test_cant_idle = test_cant_idle ||
        ((sim_clock_queue == QUEUE_LIST_END) &&             /* or clock queue empty? */
         (!(sim_asynch_enabled && sim_asynch_timer)));      /*     and not asynch? */
    #else
    test_cant_idle = test_cant_idle ||
        (sim_clock_queue == QUEUE_LIST_END);                /* or clock queue empty? */
    #endif
    test_cant_idle = test_cant_idle ||
        ((sim_clock_queue != QUEUE_LIST_END) &&             /* or clock queue not empty */
         ((sim_clock_queue->flags & UNIT_IDLE) == 0))||     /*   and event not idle-able? */
        (rtc_elapsed[tmr] < sim_idle_stable);               /* or timer not stable? */
    if (test_cant_idle) {
        if (sin_cyc)
            sim_interval = sim_interval - 1;
        return FALSE;
        }
    [...]

OR
Since the above #if conditional merely avoids two memory references, which if they're tolerable, the simplest code would be:

    t_bool sim_idle (uint32 tmr, t_bool sin_cyc)
    {
    static uint32 cyc_ms = 0;
    uint32 w_ms, w_idle, act_ms;
    int32 act_cyc;
    
    if ((!sim_idle_enab)                             ||     /* idling disabled */
        ((sim_clock_queue == QUEUE_LIST_END) &&             /* or clock queue empty? */
         (!(sim_asynch_enabled && sim_asynch_timer)))||     /*     and not asynch? */
        ((sim_clock_queue != QUEUE_LIST_END) &&             /* or clock queue not empty */
         ((sim_clock_queue->flags & UNIT_IDLE) == 0))||     /*   and event not idle-able? */
        (rtc_elapsed[tmr] < sim_idle_stable)) {             /* or timer not stable? */
        if (sin_cyc)
            sim_interval = sim_interval - 1;
        return FALSE;
        }

Once the previously mentioned async clock code is working, the #if conditional will go away.

What other "undisciplined annotations" does your tool point out?

RomeroMalaquias:
Ok,
You can see each annotation we've found here in this [repo](https://github.com/flaviommedeiros/cprojects/tree/master/patterns/list4/simh)

We have a [catalogue](http://www.dsc.ufcg.edu.br/~spg/uploads/medeiros-jucs-14.pdf) for transform undisciplined annotations to disciplined. 
This catalogue defines 9 kinds of refactoring, so the tool separates the annotations for each refactoring kind.
Each annotation is referenced in that way:
exampleN.c corresponds to name_of_file-N.c
Ex:
example1.c -> pdp18b_cpu-1.c

markpizz:
Was the change you proposed here automatically generated by a tool or was it hand crafted?

RomeroMalaquias:
Hand crafted

markpizz:
So, out of all the undisciplined annotations your system identified are there other suggested changes?

RomeroMalaquias:
No, just these.

markpizz:
Your use of "these" implies more than one suggested change, but your patch only addressed one case....

RomeroMalaquias:
Ok, i was talking about the undisciplined annotations pointed by the tool.
At this time, just this suggested change.

PR #275, opened 2016-02-09
SCP: Simh crashes if fclose(NULL) is called

mrihtar:
I remember getting this in previous versions and I fixed it locally. Now I've seen it's still here in the latest code. I can't reproduce the NULL pointer, though, but it *did* happen (under vax simulation).

### DIFF
--- a/VAX/vax730_sys.c
+++ b/VAX/vax730_sys.c
@@ -35,7 +35,7 @@
 #include "vax_defs.h"
 
 #ifdef DONT_USE_INTERNAL_ROM
-#define BOOT_CODE_FILENAME "vmb.exe"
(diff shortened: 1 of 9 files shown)


mrihtar:
I'm glad to see that my proposal to check for NULL pointer in scp.c:detach_unit() finally got through (in your commit b1318f5c 2 years later), so I guess it was not so incorrect/specific to me. :wink:

markpizz:
Notice that although the current code does "check" for a NULL pointer, is ALSO always clears the fileref pointer on ANY close.  The lack of doing this staged the need for the check in the first place.  When I fixed this, I had a real failure case that demonstrated the actual problem.  This additional logic was indeed what was missing that caused the initial problem.

Issue #292, opened 2016-03-20
Attaching ethernet to nat hangs sim when built from Homebrew


WIZARDISHUNGRY:
Testing simh on OS X; setting INCPATH and LIBPATH to point to /usr/local (Homebrew) produces binaries that hang on `attach xq0 nat:` I'm trying to get SDL Video working by linking against Homebrew. 

```
sim> attach xq0 nat:
NAT args: 
NAT network setup:
        gateway       =10.0.2.2/24(255.255.255.0)
        DNS           =10.0.2.3
        dhcp_start    =10.0.2.15
  Protocol[State]    FD  Source Address  Port   Dest. Address  Port RecvQ SendQ
Eth: opened OS device nat:
```
*(hangs)*

Builds: [simh-bad.txt](https://github.com/simh/simh/files/181566/simh-bad.txt) [simh-good.txt](https://github.com/simh/simh/files/181564/simh-good.txt)

markpizz:
Building on OS X with MacPorts is supported.  

Package managers which place results in /usr/local are indistinguishable from folks building directly from the source components of various packages.  There is a VERY DELIBERATE reason we don't want to support folks expecting arbitrary builds of personally built packages.  This reason is that when folks build things themselves there is little or no testing of exactly what they've built and as such we get build complaints from users.

The strategy that is supported is to ONLY use packages which the OS vendor supplies with their OS since the OS vendor (or the folks who maintain the package system) actually test what they're packaging and the resulting libraries, etc work reliably.

WIZARDISHUNGRY:
I appreciate the justification for not supporting an arbitrary user build. I was messing around with this hoping to get a Homebrew recipe for simh4 that works 100% correctly. Handling /usr/local like /opt/local in the makefile produces a vax binary that works for SDL graphics and nat networking.
 
Would a patch to the makefile that handles  /usr/local like /opt/local – **only** if homebrew is installed – be acceptable?

markpizz:
[...]

PR #293, opened 2016-03-21
Improve makefile for Homebrew on OS X

WIZARDISHUNGRY:
fixes simh/simh#292


markpizz:
As discussed in #292.

PR #300, opened 2016-04-07
PDP11: Record the actual source and destination values.

pkoning2:
Previously, the history would always use a register value as source or
destination as if the mode were zero, even when it wasn't.  Also, now
the destination value reflects the destination after instruction
execution rather than before.


markpizz:
This was merged manually into the master branch.

Thanks.

PR #306, opened 2016-05-08
AltairZ80: Fixed device descriptions for all devices.

psco:


markpizz:
I rebased this commit into the master branch since I'd already made and published other commits without noticing that this pull request.

Massive const changes have not been commited and pushed.  Please pickup the current codebase and possibly reclone from the github.com/simh/simh repo.

PR #310, opened 2016-05-15
VAX: replace magic numbers with constants

shattered:


markpizz:
Rebased into master branch.

Thanks.

PR #321, opened 2016-06-01
VIDEO: don't flip mouse button state while mouse moves

shattered:
This makes dragging an outline of new window work in DMD 5620 sim.


markpizz:
Hi Sergey,

Thanks for digging.

I'm not going to take a commit which merely comments out a line which is broken and probably doesn't actually do the right thing.

Does the this change to that line produce useful results for you?

    event->state = dev->state;

Let me know.

-  Mark

PR #324, opened 2016-06-03
ETHER: Fix incompatible ptr type warning and relocate _eth_get_system_id

agn453:
Compilation under Cygwin with gcc 5.3.0 - missing const gave rise to incompatible pointer type warning, and _eth_get_system_id function for _WIN32 was defined under a conditional for either _WIN32 or __CYGWIN__ and later redefined again for !_WIN32


markpizz:
Included via rebase.  Reworded the commit message to keep lines < 80

PR #349, opened 2016-11-09
BESM6: Added variable Baud rate for serial terminals.

leobru:
The Baud rate is based on emulated ticks rather than wallclock time and appears much faster than in real life.


markpizz:
Hi Leo,

The TMXR layer already has variable speed for simulated multiplexer lines.  Without this patch, a user could merely say:

    sim> ATTACH TTY port,speed=300

and things would probably work as you expected...

useful info is visible with:

    sim> SHOW MUX

If this doesn't work for you without any changes, then we can explore why in more detail and the fix the generic tmxr facilities to meet your goals.  

You might also benefit from changing:
>    /* If the TTY system is not idle, schedule the next interrupt
>     * by instruction count using the target interrupt rate of tty_rate Hz;
>     * otherwise we can wait for a roughly equivalent wallclock time period,
>     * e.g. until the next 250 Hz wallclock interrupt, but making sure 
>     * that the model time interval between GRP_SERIAL interrupts 
>     * is never less than expected.
>     * Using sim_activate_after() would be more straightforward (no need for a check
>     * as the host is faster than the target), but likely less efficient for idling.
>     */
>    if (vt_is_idle() && sim_activate_time(clocks) > 1000*MSEC/tty_rate)
>        return sim_clock_coschedule (this, tmr_poll);
>    else
>        return sim_activate(this, 1000*MSEC/tty_rate);

to

>    return sim_clock_coschedule (this, tmr_poll);

leobru:
Hi Mark,

Thank you, I'll try that tomorrow.

markpizz:
Additionally, if it happens that the serial ports on this system had programmatic port speed, the TMXR layer has APIs to explicitly set port speed on a line by line basis.  Before exploring this, getting the previously mentioned stuff working right should be the goal.  I'm just mentioning it now so you can keep it in mind.  Implementing this would certainly require more work in the device simulation since you'd have to implement whatever interface the hardware provided to perform these actions.

PR #363, opened 2016-12-16
Fix PDP-8 printing from WPS

dgesswein:


markpizz:
Thanks for this, however there are 2 Things:
1) The commit message should be worded PDP8: Fix printing from WPS
Normally I'd handle changing this myself with a rebase during the merge.
2) Your change:

        case 0:
              dev_done = dev_done | INT_LPT;                  /* set flag /
              int_req = INT_UPDATE;                           / update interrupts */
              return AC;

Starts a comment on the line which begins with dev_done and ends it on the line beginning with int_req, effectively commenting out the int_req line.  Was this intentional?

dgesswein:
On Fri, Dec 16, 2016 at 08:11:37AM -0800, Mark Pizzolato wrote:
> Thanks for this, however there are 2 Things:
> 
> Starts a comment on the line which begins with dev_done and ends it on the
> line beginning with int_req, effectively commenting out the int_req line.
>  Was this intentional?
>
Nope. I wasn't paying enough attention. I looked at the original file
which was correct. The code in the git issue has the change to the comment.
I assume it modified what I pasted but not sure. Since it was open I
copied and pasted the code from that page into the git code I cloned instead
of going back to the original file I used for initial debugging.

I tested and committed the change. Either version worked with WPS.

markpizz:
I squashed both commits into a single commit and commited the result with you as the author.

github.com/simh/simh now has a different commit sequence than your local repo and its github/dgesswein/simh clone.  You probably want drop these now and clone again next time something comes up.

BTW, I saw what you're talking about regarding the C comment looking the way it was in the issue discussion.  The asterisks were actually in the text of the discussion but got lost during the translation from 'Github Flavored Markdown' to the viewable form in the issue.  I edited your comments and merely put a blank line before the code section and the resulting stuff displayed as a code block with the asterisks in the comments all there.

PR #369, opened 2016-12-27
fixup some accumulated whitespace problems

RussellSenior:
A tree-wide whitespace fix, avoiding binary files.

Signed-off-by: Russell Senior <russell@personaltelco.net>


markpizz:
Hi Russell,

FYI: For legacy reasons, all text files in the simh codebase (.c, .h, .txt, etc) are maintained with CRLF line endings in the repo.  Source code (.c, .h) are maintained tab characters expanded to spaces so that the code looks the same for anyone who looks at it without regard to their local tab settings.  

I recommend that when using git to access/change the simh repo that your git environment be set to 'checkout as-is, commit as-is'.  I think the git setup for the line ending stuff is configured globally with 'git config --local --add core.autocrlf false'.  

There may be some cases where these general rules haven't been followed, and I'd like to specifically know about them.

Thanks.

- Mark

PR #375, opened 2017-01-16
BESM6: Improved TTY I/O (responsiveness, variable rate, turbo/authentic)

leobru:
Is this better?

Thanks, I keep forgetting about "push -f".


markpizz:
In general it is good to not know about "git push -f" since it should only be used when all interested parties agree on it and that can't be guaranteed for a public repo.

PR #376, opened 2017-01-17
BESM6: Preparing to implement punchcards.

leobru:


rcornwell:
You might want to consider using sim_card.c. I can provide a updated version for it if you want. The current version works, but has problems with binary card formats, I have fixed it.

leobru:
@rcornwell At the moment I'm going to implement the puncher only (the OS allows to represent p/c images on punch tape, so no immediate need for a p/c reader). The "binary" format I have in mind uses 3 lines of 40 Unicode characters U+2800 to U+28FF (Braille patterns) followed by a blank line to represent a card in a visible form (although not to scale). Recognizing text will be hard: the encodings (with Cyrillic extensions) could be IBM-like, CDC-like, [column-wise GOST](https://ub.fnwi.uva.nl/computermuseum//DWcodes.html#A056), [byte-wise GOST](https://ub.fnwi.uva.nl/computermuseum//DWcodes.html#A059) (horizontal, up to 120 bytes per card, with odd parity for valid bytes), so I'm not planning that at first. The puncher is intended to demonstrate the ability to punch object files in a binary format.

PR #430, opened 2017-04-07
Use correct variable name, FONTPATH instead of FONTNAME in comment

ryoon:


markpizz:
Merged with Cherry Pick to provide standard commit message format and avoid additional merge commit.

PR #433, opened 2017-04-26
VAX: Fixed Coverity issues

9track:
Fixes for coverity issues 1415454 and 1415576


9track:
Tested with VWS, UWS and DECwindows. The QVSS issue only affected polled interrupt mode, which is likely not used by most operating systems.

markpizz:
Merged with Cherry Pick

PR #381, opened 2017-01-25
PDP11 -- RX11 improvements

shattered:
I have written a RX11 simulation for MAME (using SIMH and diag listings as reference), and that simulation passes ZRXA and ZRXB tests in XXDP (SIMH's fails both).   This is a start of backport into SIMH.

PR #468, opened 2017-06-09
PDP11: Improve simulation accuracy of RX device.  

shattered:
Derived from RX11 high-level emulation in MAME (https://github.com/shattered/mame/commit/8b27f9ffb1cf542002850c0c170c980eb6ce9faa)

### DIFF
--- a/PDP11/pdp11_rx.c
+++ b/PDP11/pdp11_rx.c
@@ -50,6 +50,7 @@
 */
 
 #include "pdp11_defs.h"
+#include "sim_disk.h"
 
 #define RX_NUMTR        77                              /* tracks/disk */
 #define RX_M_TRACK      0377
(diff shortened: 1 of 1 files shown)


shattered:
any news on this one?

markpizz:
Well, this is hung up due to the fact that it is a single commit with a large number of changes.

The large number of changes mix many cosmetic changes together with substantive changes.  The cosmetic changes (using symbols instead of hard coded numeric values, adding debug output, etc.).

This along with the rather terse single line commit message with no related detailed explanation of the specific problems addressed and why they needed to be addressed, makes this a significant project for someone else to verify.  When changes are made to modules written by Bob Supnik, he gets first crack at reviewing the thinking and implementation of such changes.  In general, he would never even look at a large change set like this.  After extensive discussion,  he might be convinced to embrace the need for a change.  I'd suggest that you rework this change set isolating into separate commits the cosmetic from the functional changes and then explain deeply the justification for the functional changes.  It would be good if you could precisely spell out how to demonstrate the diagnostic tests that failed (providing attached files (in a zip container) to the related discussion so that someone reviewing it can demonstrate the problem themselves).

shattered:
No problem, and could you write all this up and publish on Github as contribution guidelines for this repo?

+ maybe I should reopen older https://github.com/simh/simh/pull/381 -- it has smaller set of changes

PR #485, opened 2017-10-16
Add LIBPATH checks to support libraries installed via Homebrew

redshirtrob:
Update the README with directions for installing the missing
libraries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -420,6 +420,11 @@ The MacPorts package manager is available to provide these external packages.  O
     # port install vde2
     # port install libsdl2
 
+Or, the Homebrew package manager is available to provide these external packages.  Once Homebrew is installed, these commands will install the required dependent packages:
+
+    # brew install vde
+    # brew install sdl2
+    
 ###### Linux - Dependencies
 
 Different Linux distributions have different package managment systems:
--- a/makefile
+++ b/makefile
@@ -249,6 +249,10 @@ ifeq ($(WIN32),)  #*nix Environments (&& cygwin)
         LIBPATH += /opt/local/lib
         OS_LDFLAGS += -L/opt/local/lib
       endif
+      ifeq (libopt,$(shell if $(TEST) -d /usr/local/lib; then echo libopt; fi))
+        LIBPATH += /usr/local/lib
+        OS_LDFLAGS += -L/usr/local/lib
+      endif
       ifeq (libXt,$(shell if $(TEST) -d /usr/X11/lib; then echo libXt; fi))
         LIBPATH += /usr/X11/lib
         OS_LDFLAGS += -L/usr/X11/lib


markpizz:
Sorry.  Using Homebrew libraries on OS X isn't supported.  We explicitly support MacPorts and really aren't interested in the expanded complexity matrix that exists when other package sources come into the picture.

redshirtrob:
What a massive disappointment...

markpizz:
There are good number of headaches that will come along with just picking up things found in /usr/local/lib, which is where Homebrew puts/keeps things, since this is also where most 'build it yourself' packages will also land.  Having a contained package comes with the significant value that those actually packaging things actually test the resultant packages and make sure that things actually work.  When folks build their own libraries from scratch, we've seen a good number of subtle problems which we're not really interested in fielding.

redshirtrob:
I guess I'm surprised there are that many people actually installing software in /usr/local on macOS, outside of Homebrew. Given the increasing popularity of Homebrew over Fink and MacPorts in the past 8 years, I expected a warmer reception. Nevertheless, I have no pressing need to build this software on macOS, I was merely trying to upstream something I thought might be helpful to others. 

I realize there's a philosophical issue with where Homebrew places things. I might be persuaded to disagree if it was anything resembling a proper Unix, but I prefer to be pragmatic. But it's not my project to maintain so I understand your position.

markpizz:
I think we're now allowing both MacPorts and HomeBrew on macOS. Your comments and those of others inspired a way to tolerate HomeBrew's use of /usr/local while still 'mostly' avoiding complications with user built libraries.

jguillaumes:
I moved to Homebrew some months ago. I patched the makefile to work with /usr/local and found the only thing to be modified is the setup of LIBPATH, since /usr/local/include is already taken (it comes from the compiler configuration). Alternatively, a symlink between /usr/local and /opt/local is a patchy and ugly solution (that works).

I think you can find TWO possible problems related to homebrew:

1) The posible simultaneous use of macports and homebrew, That is NOT recommended by the homebrew guys and imho should be tagged as unsupported.

2) The “Cellar only” homebrew packages. Those do not create symlinks in /usr/local/bin nor /usr/local/lib et al; homebrew installs them in /usr/local/Cellar/package/… and warns the user to define LDFLAGS and CPPFLAGS to use the brewed packages in build processes. That happens with packages that can conflict with Apple provided ones; the best example being openssl. 

More posible pain points: some packages have different configuration in macports and homebrew. VDE, for example, is provided without tun tap support in the homebrew version. The macports version can be installed with tun tap support adding “+tuntap”, while the brew version must be compiled from sources after modifying the “formula” to incorporate that.

In general, after some initial suffering I’m completely OK with homebrew, and the changes needed to support it in the makefile are quite small.

markpizz:
Please try the latest commit.

jguillaumes:
Wow! Now it adds the whole Cellar to the build! :)

```
$ ./BIN/vax

MicroVAX 3900 simulator V4.0-0 Beta        git commit id: 0da9a49c
sim> show version
MicroVAX 3900 simulator V4.0-0 Beta
    Simulator Framework Capabilities:
        64b data
        64b addresses
        Threaded Ethernet Packet transports:PCAP:TAP:VDE:NAT:UDP
        Idle/Throttling support is available
        Virtual Hard Disk (VHD) support
        Asynchronous I/O support (Lock free asynchronous event queue)
        Asynchronous Clock support
        FrontPanel API Version 6
    Host Platform:
        Compiler: GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.38)
        Simulator Compiled as C arch: x64 (Release Build) on Dec  3 2017 at 01:45:23
        Memory Access: Little Endian
        Memory Pointer Size: 64 bits
        Large File (>2GB) support
        SDL Video support: SDL Version 2.0.7
        PCRE RegEx support for EXPECT commands
        OS clock resolution: 1ms
        Time taken by msleep(1): 1ms
        OS: Darwin macjordie.jguillaumes.dyndns.org 17.2.0 Darwin Kernel Version 17.2.0: Fri Sep 29 18:27:05 PDT 2017; root:xnu-4570.20.62~3/RELEASE_X86_64 x86_64
        git commit id: 0da9a49c
sim> show ether
ETH devices:
 eth0	tap:tapN                             (Integrated Tun/Tap support)
 eth1	vde:device                           (Integrated VDE support)
 eth2	nat:{optional-nat-parameters}        (Integrated NAT (SLiRP) support)
 eth3	udp:sourceport:remotehost:remoteport (Integrated UDP bridge support)
libpcap version 1.8.1 -- Apple version 79.20.1
sim> 
```
Awesome!

PR #515, opened 2017-12-29
I7000: Fixed coverity warnings.

rcornwell:


markpizz:
You 'fixed' the warnings that coverity is issuing about not checking the return status when you call sim_process_event(), but merely explicitly ignoring all status might not be the right answer.  My first thought was that a console user types ^E and that is detected by a keyboard probe of the console device.  As it turns out, that won't be a problem since those probes return SCPE_OK and set the global stop_cpu to 1 if ^E is observed, which your sim_instr() instruction loop should stop on.  There might be other cases though...

rcornwell:
In both of the cases I ignore the return code, the simulator will return to the command prompt after it has processed all pending I/O. All I need do is let any events that need to be processes to be scheduled. So it does not matter what is returned.

PR #522, opened 2018-01-26
sim_ether.c: Add suppor to attach to specific VDE virtual switch port

jguillaumes:
# Rationale

VDE implements a virtual switch which allows buiding virtual networks for  virtual machines and simulators. The implemented switch is actually a *managed* switch, capable of, for instance, defining virtual ports as tagged for VLAN definition.

This change allows the simh user to specify a specific virtual port to attach the simulated ethernet card to. That would allow using VLAN tagged networks to separate -for example- the testing machines in a virtual network.

# User impact

The user can add an optional port number to the VDE attach string. For example:

```
attach xq vde:/tmp/vde.ctl:8
```

That would attach the XQ device to the virtual port number 8. That port has to be previously defined using vdeterm and set up correspondlingly.

The port number is optional. If it is not specified simh will attach to whatever port VDE makes available.

### DIFF
--- a/sim_ether.c
+++ b/sim_ether.c
@@ -2084,6 +2084,11 @@ if (0 == strncmp("tap:", savname, 4)) {
 else { /* !tap: */
   if (0 == strncmp("vde:", savname, 4)) {
 #if defined(HAVE_VDE_NETWORK)
+    int vdeport = 0;            /* VDE port. zero => assign automatically */
+    char *vdeschem_s = NULL;    /* VDE device scheme ("vde:") */
+    char *vdeswitch_s = NULL;   /* VDE swtich name */
+    char *vdeport_s = NULL;     /* VDE switch port (optional), numeric */
+      
     struct vde_open_args voa;
     const char *devname = savname + 4;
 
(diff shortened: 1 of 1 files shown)


sim_ether.c
```diff
@@ -2083,7 +2083,12 @@ if (0 == strncmp("tap:", savname, 4)) {
   }
 else { /* !tap: */
   if (0 == strncmp("vde:", savname, 4)) {
-#if defined(HAVE_VDE_NETWORK)
+#if defined(HAVE_VDE_NETWORK)
+    int vdeport = 0;            /* VDE port. zero => assign automatically */
+    char *vdeschem_s = NULL;    /* VDE device scheme ("vde:") */
+    char *vdeswitch_s = NULL;   /* VDE swtich name */
```
tswachter: /* VDE swtich name */ should be /* VDE switch name */

Rhialto:
@jguillaumes Good idea for klh10 too!

jguillaumes:
That's next in my to-do list. :)

markpizz:
Great idea.

The implementation had a few minor issues:
* mixed line endings in sim_ether.c
* overwriting of the device name string while parsing it (it might be needed later), so its format can be corrected, but the changed result must be a valid input for a subsequent call with the updated value
* missed mention of the vde switch port in the output of SHOW ETHERNET
* undesirable merge commits prior to pull request - cherry picked changes
* committed changes with Jordi as author.

markpizz:
Sorry I didn't push the change out before I closed this issue.  A git glitch on my part.  :-(

It is there now.

Thanks again.

PR #526, opened 2018-02-17
DISPLAY: Update display code to support DEC Type 340, and 36 switches

philbudne:
Only interface code to Type 340 is for Richard Cornwell's KA10
(but could be used on PDP-1/4/7/9 as well)


larsbrinkhoff:
Any known caveats?

philbudne:
1.  I just (force) pushed an update to my 340 branch,
    I had left out updates to PDP11/pdp11_vt.c

2.  PDP1/spacewar1/spacewar.rim behaves exactly the
    same with and without my changes:
    It displays a frozen screen.

    I've never tried it since the merge of my display code,
    with sim_video/SDL or X11, so I can't say where the issue
    might be....

BUT display.c works using X11 and the "Type 340" display system under
Richard Cornwell's KA10 simulation.

markpizz:
I don't know how to play the spacewar game, but as far as I can tell, building the pdp1 simulator with OS provided SDL2 package on Ubuntu produces a working program.  I follow the  simple instructions in the pdp1/spacewar1/README file and:
```
sim> set dpy enable
sim> attach ptr spacewar.rim
sim> boot ptr
```
The build on Windows also produces a program which behaves the same. 
With the program running, pressing keys seem to change the behavior of the elements, so the pieces seem to be working.  

As for this commit.  A compile on OS X produces a compiler warning:
```
display/display.c:1004:35: warning: '/*' within block comment [-Wcomment]
    /*printf("down '%c'\r\n", k); /**/
```

Additionally, you've changed the call signatures for set_cpu_switches() and get_cpu_switches().  There are other modules in the display directory which reference these APIs:
- vttest.c
- xy.c
- test.c
It would seem logical if all were changed to use the revised interfaces, right?

philbudne:
Yes: I'll fix vttest.c (which I found broken by a signature change to
display_init) and test.c (tho it's doubtful anyone has built/used them
since I wrote the code 15 years ago!) and make sure they include
display.h so they won't build if the signatures change!

xy.c seems to be an unused, older copy of display.c; I have some
recall of a file being renamed for some reason or another when it was
pulled into the SimH repo.

markpizz:
Ahhhh...  I think I copied xy.c to display.c to avoid a file name conflict.  I left the original file in the event that some external (non simh code) referenced things by the original name.  It is fine by me if it goes away.

PR #538, opened 2018-03-15
Continuous build in Travis CI.

larsbrinkhoff:
This adds configuration for building in Travis CI.  For this to take effect, someone will have to login to Travis CI and enable building for the simh repository.


larsbrinkhoff:
I also tried building in AppVeyour, but I couldn't make it work.  MSBuild wouldn't accept the the old project and solution files.  And when I tried `/toolsversion:3.5` that didn't work either.

The attempts are visible here:  
https://ci.appveyor.com/project/larsbrinkhoff/simh/history

larsbrinkhoff:
This just runs `make all`.  It would also be possible to build all simulators separately in parallel.

larsbrinkhoff:
@markpizz Please also login to Travis, then go here https://travis-ci.org/profile/markpizz to enable the build.

larsbrinkhoff:
Thanks.  The next time someone does a push, a build will be started.

markpizz:
Thanks for setting this up so far!

I looked at changing the .travis.yml so that all the potentially useful prerequisite packages are installed:  
For Linux the log shows that it would like: libsdl2-dev, libpcap-dev and libvdeplug-dev
For OSX the log shows that it would like: sdl2 and vde

I saw some documentation about how to install prerequisite packages, but not how to describe different steps for different host platforms.  Can you look into this?

As for attempting similar results for a Windows build, there are several things to be addressed:
1) The current projects/solution are Visual Studio 2008 projects which will auto-convert to a newer version when a newer Visual Studio GUI opens the solution file.  Thus all versions from 2008 onward are supported (*).  Meanwhile, when actually scripting from the solution/project definitions, MSBuild came along with a later version of Visual Studio (prior versions used vcbuild).
2) The actual build has various external dependencies which are all required and precompiled working versions are contained in a windows-build repo/directory which needs to be located right next to the directory containing the simh repo.  It is not clear how to  get this available and located appropriately, but point 1 needs to be addressed first anyway.

\* The pre-compiled windows-build components are usable for all Visual Studio versions up through, but not including Visual Studio Community 2017.  There are many Visual Studio variations here and in general things are forward compatible, but not backward compatible (i.e. VS 2013 projects aren't useful on earlier VS version (VS2008 especially)).

larsbrinkhoff:
Travis will set the environment variable TRAVIS_OS_NAME to "linux" or "osx".  I have this handled in another project, so I can look into it tomorrow.

larsbrinkhoff:
Prerequisite librarires are adressed by #540.

markpizz:
I believe that there is great benefit to running building and testing on Linux and OS/X.  Essentially all of the substantive issues that may be flagged on Linux and OS/X produce indirect benefits on Windows, without any active CI build/test on Windows.

If you're really driven to get Windows working, you should personally get the details for Windows working on a local Windows VM.  You don't need a formal license for Windows to do this.  You can install a current version of Windows (for testing) and merely ignore the on-screen message that suggests that Windows isn't properly licensed.  Testing like this is within the allowed usage scenarios of a test windows install.

More said in #746

larsbrinkhoff:
In my view, things that are not subject to automated building and testing are easily subject to bitrot.  If it's not a huge effort, I like to make use of the available CI services.  Running a simple build script is usually quite easy.  Testing may be harder.

PR #540, opened 2018-03-19
Install prerequisite libraries for Travis CI.

larsbrinkhoff:
This install some prerequisite  libraries for the Travis CI build.

For an unknown reason I had to install libegl1-mesa-dev and libgles2-mesa-dev too to make apt happy.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,4 +2,6 @@ os:
   - linux
   - osx
 language: c
+sudo: required
+install: sh -ex .travis/deps.sh
 script: make all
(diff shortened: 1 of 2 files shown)


markpizz:
How about locating the deps.sh script in a .travis directory?
It would be nice if the .travis.yml could be there as well (like the .github directory).  Oh well.

larsbrinkhoff:
Sure, I moved it.  I don't think the yml file can be moved though.

markpizz:
Also, please add libsdl2-ttf-dev to the Linux package set.  No such package is yet available for HomeBrew on OS X.

larsbrinkhoff:
Done.

PR #541, opened 2018-03-19
Install sdl2_ttf for Travis macOS.

larsbrinkhoff:
As discussed on the mailing list.


markpizz:
Lars,

This change was obvious given your initial setup, so I'd committed that along with a makefile cleanup (suggesting the brew install when the ttf support isn't found).  That commit and this pull request overlapped.

PR #544, opened 2018-03-23
AltairZ80: Set conditional under CYGWIN so the SIMH device works

agn453:
Under CYGWIN the conditional tests in the altairz80_sio.c source file for the SIMH pseudo device's getHostsFilenamesCmd and resetSIMHInterfaceCmd functions need to have UNIX_PLATFORM set to 1 - otherwise they do nothing! With this fix, the HDIR host command from CP/M now displays a list of files in host system's current directory under CYGWIN.


markpizz:
Well, I used cherry pick to get your commit message and the change.  I then committed that change with you as the author.  This was necessary to avoid the additional 24 merge commits what were part of this pull request.  If you're not doing other simh work in your local or github repo, this would be a time to delete the github and local repos and clone things again from github.com/simh/simh.

PR #556, opened 2018-05-06
I650: Fix Coverity Warning

rsanchovilla:


markpizz:
Roberto,

This change:

```
            // should wait for drum to fetch data?
-            if ((bReadData) && (AR >= 0) && (AR < (int)MEMSIZE)) {
+            if ((bReadData) && (AR < (int)MEMSIZE)) {
                if ((AR % 50) != DrumAddr) continue; // yes
            }
            MachineCycle = 3; // exec instr
```

 DOES fix the specific Coverity warning about the part of that comparison expression which is a wasted test (unsigned >= 0).  However, with that removed, the remaining test ends up now being a signed/unsigned comparison.  Avoid the cast (int) of MEMSIZE would seem to be a complete fix.

This would be an opportunity to unwind this commit and do it better.

- Mark

rsanchovilla:
Fixed.

PR #561, opened 2018-05-13
KA10: Fix 340 bug.

larsbrinkhoff:
In the `vector` function, the `flags` variable isn't initialized.  The function can sometimes return boolean true even though it shouldn't.  This causes the 340 to escape vector mode, with display list havoc as a result.


larsbrinkhoff:
Thanks!

CC @rcornwell, my 340 bug fix is in.

PR #571, opened 2018-05-18
I650: Release 3

rsanchovilla:
New Hardware support:
- IBM 653 Storage Unit: provides machine opcodes
for Floating Point, Immediate Access Storage (IAS),
Three Index registers, Cards Punch-read
synchronizers 2 and 3.
New Software included:
- FORTRANSIT: version II (S), plus run time PACKAGE
with standard Fortran functions.
- Reorganized sw directory, separating each language
in its own folder. Each one Includes a 00_readme.txt
file with restoration notes and comments.
New features:
- Support for SOAP opcode mnemonics in addition
to regular IBM mnemonics
- FAST / REALTIME CPU options
- PROP pseudo register
- CARDDECK ECHOLAST command

### DIFF
--- a/scp.c
+++ b/scp.c
@@ -3291,7 +3291,7 @@ strlcpy (abuf, fcptr, sizeof(abuf));
 c = abuf;
 do_arg[10] = NULL;                                      /* make sure the argument list always ends with a NULL */
 for (nargs = 0; nargs < 10; ) {                         /* extract arguments */
-    while (sim_isspace (*c))                                /* skip blanks */
+    while (sim_isspace (*c))                            /* skip blanks */
         c++;
     if (*c == 0)                                        /* all done? */
         do_arg [nargs++] = NULL;                        /* null argument */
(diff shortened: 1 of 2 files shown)


markpizz:
The earlier compile warnings are now fixed.
Meanwhile, something you changed today has now added tab characters to i650_cdr.c which weren't there in your earlier version or I would have mentioned it earlier.  

You really won't have to deal with this if you change the Visual Studio editor for .c and .h files to insert spaces for tabs!!!!!

Please undo and fix and go again.  You can wait until tomorrow to come back to this.

markpizz:
You should be able to use Visual Studio to search for tab characters as follows.  I'm doing this on Visual Studio 2008, so if you're using that, it will be precise, if you're using a later version it will be similar:

Menu->Edit->Find In Files
Set "Find What:" to \t
Set "Look in:" to the path to your I650 folder
Check "include sub-folders"
Set Find options "Use:" to "Regular Expressions"
Click on "Find All"

What I'm seeing here shows i650_cdr.c lines 561 thru 572 having some tab characters.

PR #591, opened 2018-06-20
display: Fix X11 crash when DISPLAY unset.

larsbrinkhoff:
If the DISPLAY environment variable isn't set, the ws_init call to XkbSetDetectableAutoRepeat will crash due to a NULL pointer.

### DIFF
--- a/display/x11.c
+++ b/display/x11.c
@@ -272,6 +272,10 @@ ws_init(const char *crtname,    /* crt type name */
     xpixels = xp;               /* save screen size */
     ypixels = yp;
 
+    if (getenv ("DISPLAY") == NULL) {
+        return 0;
+    }
+
     XtToolkitInitialize();
     app_context = XtCreateApplicationContext();
     argc = 0;


markpizz:
This change is fine.

Meanwhile, be aware, that direct use of the x11 (and w32) interface layer code is deprecated for simh projects and will not be supported in the future.  The SDL functionality provided by sim_ws and sim_video will be the future.  A significant change to sim_video and other related display support is pending detailed review and adjustment and should be available 'soon'.

larsbrinkhoff:
Thanks.

I'm primarily fixing things in the existing Type 340 simulation.  Will it be moved over to the new APIs?

markpizz:
A natural part of the merge will be to support the 340.  The 340 stuff lives on top of display.c, right?  If so, then it shouldn't be a problem.  You'll probably have to be involved in testing though.

larsbrinkhoff:
Yes, I think that's right.

Sure, I'll do the testing.

I think @philbudne should be in the loop on this.

PR #599, opened 2018-07-28
PDP11: Rubin 10-11 interface

larsbrinkhoff:
### Context

The MIT AI lab PDP-10 had a device to access the core memories of up to eight attached PDP-11s, called the Rubin 10-11 interface.  The PDP-11 machines were used as peripheral processors or used shared memory to talk to ITS on the PDP-10.  The principle of operation is quite simple.  The PDP-10 can read or write individual words using Unibus.  The PDP-11s can only access its own memory, not the PDP-10 memory.

I have added this capability to the SIMH PDP-11 simulator.  It uses the socket API to connect to a PDP-10 simulator.  The PDP-10 can send read and write messages, and the PDP-11 will reply with a status code and data for reads.

This is only at a proof-of-concept stage.  The code is most likely not acceptable for merging yet.  Since this is implemented as point-to-point message passing, maybe TMXR can be used instead of sockets.

My question is, is this feature something that could be considered for inclusion?  If so, I could post what I have and get comments on how to rewrite it more properly.


larsbrinkhoff:
Thanks.  I posted my code as a pull request here.  (I can either continue to rework it in this pull request, or close it and open another one later.)

This is still only prototype stage.  Work in progress.

markpizz:
You've only got one side of the activity provided here, but there are a bunch of problems with this:
1) As has already been mentioned, direct use of sockets.  More problematically than was expected are direct (and includes) of platform specific socket code, rather than using the sim_sock portable socket layer which also shouldn't be used in simulator and/or device simulator code.
2) There is nothing which leverages the simh DEVICE and event scheduling model going on here.  Everything has been done with direct changes to the the PDP11 CPU module.  Absolutely no changes to pdp11_cpu.c should be required to add the functionality you've described.  A module named something like pdp11_ten11.c which defines a simh DEVICE (possibly named TEN11) with a couple of additional lines in pdp11_sys.c that add a reference to this DEVICE to the sim_devices array should be the only required changes to existing code in this simulator.
3) The TEN11 device should have an attach routine which takes arguments that describe who/what/how this side will be communicating with something else.  The attach routine will schedule a unit that will have its service routine called after some number of instructions (possibly 1).  The service routine will gather and execute commands from the requesting PDP10 system pushing back responses and reschedule itself to run after some number of instructions.  Notice that nothing I've said so far mentions any details about how data is exchanged with the PDP10 system (sockets, TMXR, shared memory).
4) The code provided in your pull request uses the CPU's deposit and examine interfaces.  These interfaces were never intended to be used by devices, but specifically are part of the infrastructure that provides the simh command interface.  These capabilities are a distinct superset of the described requirements to merely read or write words in the processor's memory.  Memory is directly accessible to devices (that may be doing DMA) by referencing elements of the array "uint16 *M;".  The upper bound on the array size is MEMSIZE.  The CPU's I/O page and all of its memory mapped consequences are directly available through the CPU's examine and deposit interfaces.

I've put together the PDP11 side of what a TEN11 device should be (leveraging what you provided).  This device has two modes: 1) the NETWORK mode which uses TMXR to detail with connection and data delivery and 2) A shared memory mode which puts the PDP11's memory in a shared memory region which can be directly accessed by another simulator in the same system.  The TMXR model uses the same wire protocol (slightly extended when an error occurs) but the messages are framed by the TMXR's default packet framing around the provided messages.  The message structure that you used 'could' be leveraged directly by the TMXR model, BUT that would be more work given that both sides of this conversation should be using TMXR, the existing message framing comes for free.  This is visible in the lars/ten11 branch of https://github.com/markpizz/simh.  Since I didn't have both sides of this dialog, this hasn't been tested, but it should be pretty close for this side.

larsbrinkhoff:
Thank you very much!  I will test this.  The other side is in @rcornwell's KA10 simulator; my changes are available from my GitHub fork.

Some follow-up questions:

1. Is it possible to do poll() or select() from the simulator main loop?  Talking across a network connection depends on low latency.  My prototype checked for input between every instruction, but that wasn't what I had in mind for the finished solution.

2. The external interface can access the Unibus I/O space in addition to core memory.  Sorry, I didn't make that clear.  That's why it seemed convenient to reuse cpu_ex and cpu_dep which already had logic for this.  Is there some other function to use instead?  If not, of course I can duplicate that logic if that's the right thing to go about it.

I will build on what you posted.  I see no need for double framing, for example.

markpizz:
1) Change the TEN11_POLL #define or DEPOSIT TEN11 POLL n in a configuration file.  This will cause exactly what you want to happen with the frequency you desire.  If 1 is the correct value for best system behavior, that is fine.
2) OK, then using the cpu device's deposit and examine routines is fine.

There really isn't any negative consequence for the double framing.  Your build() routine works well to build the packet with the length prefix which is readily handed directly for transmission.  The consequence is that the packet on the wire is two bytes larger.  The overhead of processing any packet is far more consequential than an extra two bytes of payload in a short message.

Since I didn't have both sides to test, I didn't end up having anything which confirms that a transmitted packet has made it on the wire (checking the status from tmxr_put_packet_ln()).  There are two possible cases: 1) the previous transmitted packet hasn't been fully transmitted SCPE_STALL (this case is impossible given the single request-response protocol), and 2) the TCP session is reset or otherwise closed just before the transmit and the transmit fails SCPE_LOST (for the PDP11 side of this, that problem is a don't care.  The next time the the connection is established by either side, things will work again).

larsbrinkhoff:
Hello,

I'm now ready to try to test this.  Do I understand it correctly that a network mode connection should be configured like this?

```
set ten11 enable
set ten11 mode=network
set ten11 peer=localhost:1235
attach ten11 1234
```

As far as I understand, this only listens to a port.  Is it possible to use your code to make a connection to a port?  The other side (@rcornwell's KA10 simulator) should also use TMXR, so one or the other has to make the connection.

markpizz:
That configuration would work fine (I don't remember the default mode, so the mode line might not be necessary).

The setup will dynamically establish the connection in both directions.  By that, I mean that both sides will try to establish a connection and both sides will be listening for one.  The first successful attempt will be used, and when a connection drops (due to one end disappearing), it will be reestablished whenever the peer side returns to availability.

The PDP10 side of this is the more complicated device since there are N potential links to pdp11_ten11 devices.   Were I writing this, there would be one device with N units that could each be configured with a peer and could be attached.  Each of the N would map to a TMXR line structure with its own listen port and peer description.

markpizz:
> Each of the N would map to a TMXR line structure with its own listen port and peer description.

The TMXR line structure is a TMLN.

I might use 2 additional UNIT_DIS UNIT structures to actually do the input/output and connection polling.  The initial N units would merely be the ATTACH and configuration targets.

markpizz:
You could probably prove the functionality with your existing implementation just by wrapping the packets you put on the wire with a 2 byte length (around the existing 1 byte length) and stripping that for the response data.

larsbrinkhoff:
I have been struggling to make this work, but now I think I'm onto something.

Does it seem right that I need to add `ten11_ldsc.rcve = 1` and `tmxr_poll_rx()`?

larsbrinkhoff:
It seems to mostly work now.  But suddenly, the PDP-11 sends this to the TMXR line:

    Disconnected from the PDP-11 simulator

larsbrinkhoff:
Never mind that last comment!

larsbrinkhoff:
It seems to mostly work now, but suddenly hangs for some reason that I haven't figured out yet.

CC @aap who is responsible for implementing the protocol on the Knight TV.  That's a PDP-11/10 with bitmapped frame buffers on Intel 1103 RAM chips.

larsbrinkhoff:
There is something strange going on.  Both sides are happily exchanging messages for a while, but then it stops.

SIMH calls send() in `sim_write_sock` which claims data was sent.  However read() on the other side doesn't return.  I run tcpdump and don't see any packet.

This is the last exchange.  The PDP-10 side first sends two bytes with the TMXR packet length.  Then 7 more with a DATO request.  SIMH sees it and calls `tmxr_put_packet_ln`, utimately calling send().  But there's no payload on the wire, just an ACK.

Port 1234 is the PDP-10 side, and 54138 is the SIMH side.

```
14:47:48.685363 IP localhost.1234 > localhost.54138: Flags [P.], seq 214:216, ack 122, win 342, options [nop,nop,TS val 782296757 ecr 782296757], length 2
        0x0000:  4500 0036 e638 4000 4006 5687 7f00 0001
        0x0010:  7f00 0001 04d2 d37a 1cc7 46c2 0845 34fa
        0x0020:  8018 0156 fe2a 0000 0101 080a 2ea0 e6b5
        0x0030:  2ea0 e6b5 0007
14:47:48.685419 IP localhost.1234 > localhost.54138: Flags [P.], seq 216:223, ack 122, win 342, options [nop,nop,TS val 782296757 ecr 782296757], length 7
        0x0000:  4500 003b e639 4000 4006 5681 7f00 0001
        0x0010:  7f00 0001 04d2 d37a 1cc7 46c4 0845 34fa
        0x0020:  8018 0156 fe2f 0000 0101 080a 2ea0 e6b5
        0x0030:  2ea0 e6b5 0601 0001 001f 0a
14:47:48.685432 IP localhost.54138 > localhost.1234: Flags [.], ack 223, win 342, options [nop,nop,TS val 782296757 ecr 782296757], length 0
        0x0000:  4500 0034 97a4 4000 4006 a51d 7f00 0001
        0x0010:  7f00 0001 d37a 04d2 0845 34fa 1cc7 46cb
        0x0020:  8010 0156 fe28 0000 0101 080a 2ea0 e6b5
        0x0030:  2ea0 e6b5

markpizz:
Maybe if I could see what you're actually doing I might have something specific to say.

Apart from showing your code, there may be a very different way to solve what you are trying to solve.

I suggest this since I really cringe at the amount of OS overhead needed to achieve the small change in the memory contents of one simulator by another simulator using any socket based approach to solve this.  I say 'any' since the OS overhead is similar with the direct socket calls in your original approach vs the approach using TMXR.  Compare this overhead to the almost 0 cost using a shared memory approach with both simulators running on the same host.   

I'm thinking that you could run a handful of PDP11 simulators with each using simh throttling to some execution rate similar to the original hardware execution rate along with the PDP10 simulator on almost any multi-core host system and things would run reasonably well if all of the changes you're modeling was done in shared memory.  At first glance this doesn't meet your additional goal of having a front panel for each of the simulated systems that also modeled the original hardware.  This completely independent problem should be somewhat easy to solve if the PDP11 and the PDP10 simulator used the sim_frontpanel API to drive the frontpanel display and control activities of the simulator(s).  The current sim_frontpanel API doesn't yet support having the frontpanel application on a different host from the simulator it is controlling, but this execution model was envisioned from the beginning, but actual support was deferred until demand existed.  A frontpanel application (or applications for different simulators) can be implemented now with the existing local API and once that is working, the only change needed would be the startup API which initializes the interface to the simulator.  If you go down this path, I'll be glad to provide support for simulators remote from the front panel application.

markpizz:
What is on the other side of the connection?

Even with that case, the example I provided for the PDP11 side of the TMXR based connection, I provided both a TCP and a shared memory mode for the data exchange.  That model could readily be implemented for the PDP10 side (with multiple connections) with each connection configured as either TCP or shared memory.

aap:
The other side are my TV11 emulator for the Knight TV system: https://github.com/aap/tv11
and my PDP-6 emulator for the 10-6 connection which is using almost the same protocol: https://github.com/aap/pdp6

I haven't actually looked at your code and left that to lars. I'm just bugging him for the last couple of day to implement the final protocol.

larsbrinkhoff:
This is my code at this point:
- https://github.com/larsbrinkhoff/simh/blob/mark/ten11/PDP11/pdp11_ten11.c
- https://github.com/larsbrinkhoff/ka10-simh/blob/lars/unibus2/PDP10/ka10_ten11.c

I can provide disk images for the ka10 simulator if necessary.

At this stage I only tried to add to the ka10 ten11 the TMXR two-byte length prefix to the messages.  The next step would be to rewrite it to use proper SIMH devices, units, service routines, etc.

larsbrinkhoff:
I understand the overhead is cringe worthy, but I'd like to be able to use a PiDP-11 with an rPi built in, hooked up to a PC or the future PiDP-10.  If SIMH supported using select() or similar instead of frequent polling, the overhead would be smaller.

Keep in mind some applications need to access the Unibus I/O page.  I believe the shared memory option doesn't support that, correct?

markpizz:
> I understand the overhead is cringe worthy, but I'd like to be able to use a PiDP-11 with an rPi built in, hooked up to a PC or the future PiDP-10. 

If the PiDP-11 used the sim_frontpanel API to interface between the simulator and the physical frontpanel hardware (or GUI implementation) and the simulator, then, as I said, the only change would be changing the frontpanel startup API to the as yet unimplemented remote simulator startup.

> If SIMH supported using select() or similar instead of frequent polling, the overhead would be smaller.

The overhead would be smaller, but still at least one system calls per instruction executed.

> Keep in mind some applications need to access the Unibus I/O page. I believe the shared memory option doesn't support that, correct?

That is currently true, but I didn't know about the I/O page access requirement when I first implemented it.  The current implementation could be enhanced to support I/O page access and the result could easily avoid system calls.

larsbrinkhoff:
I think @aap and I are coming around to agree with @markpizz about using shared memory.

First, experience with our socket-based (through TMXR) shared memory has showed that using it locally inside one machine is OK, but across Ethernet is slow.  It may be OK for some use cases, such as those computers that only used the shared memory interface to load programs occasionally.

Second, we have some other shared memory interfaces in mind for the future.  The MIT KL10 had two of those: DTE20 and DL10.  Unlike Rubin's 10-11 interface, they are two-way: the PDP-11 can access all of the PDP-10 memory, and vice versa.  Solving this is a different problem, and we believe using shared memory on the host is the best way.

PR #602, opened 2018-07-28
PDP11: Allow setting core size to 8K, 24K, 40K, and 56K.

larsbrinkhoff:
I have some historical software that claims some PDP-11s had 24K and 56K octets of core memory.

This adds the full set of 8K increments < 64K.


markpizz:
Merged with Cherry pick to keep commit sequence linear.

PR #611, opened 2018-08-19
CIO memory overwrite fix and warning cleanup

sethm:
This pull request contains two changes:

1. Fixes a bug that caused 12 bytes of SVR3 kernel memory to be overwritten whenever the PORTS or CTC card was initialized.

2. Cleans up all warnings in 3B2 code generated when compiled with `-Wall`, and moves static declarations out of header files and into corresponding source files.


sethm:
Please note: My Windows development environment is still not set up, so I have not compiled under Windows. I'll be working on that today. I have cleared all warnings issued when compiled with clang on OS X using `-Wall`, so I have high hopes that there will be no warnings under Windows.

PR #614, opened 2018-09-04
PDP11: ROM device.

larsbrinkhoff:
This is a ROM device for booting from the VT11 card in a GT40.

I'm posting this for review.  This is in very raw form.  If this approach is OK, I'll rework it as necessary for inclusion in SIMH.

### DIFF
--- a/PDP11/pdp11_rom.c
+++ b/PDP11/pdp11_rom.c
@@ -1,0 +1,221 @@
+/* pdp11_rom.c: Read-Only Memory.
+
+   Copyright (c) 2018, Lars Brinkhoff
+
+   Permission is hereby granted, free of charge, to any person obtaining a
+   copy of this software and associated documentation files (the "Software"),
+   to deal in the Software without restriction, including without limitation
+   the rights to use, copy, modify, merge, publish, distribute, sublicense,
+   and/or sell copies of the Software, and to permit persons to whom the
+   Software is furnished to do so, subject to the following conditions:
+
+   The above copyright notice and this permission notice shall be included in
+   all copies or substantial portions of the Software.
+
+   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+   IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+   FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+   LARS BRINKHOFF BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+   IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+   CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
+
+#include "pdp11_defs.h"
+
+t_stat rom_ex (t_value *vptr, t_addr addr, UNIT *uptr, int32 sw);
+t_stat rom_dep (t_value val, t_addr addr, UNIT *uptr, int32 sw);
+t_stat rom_rd (int32 *data, int32 PA, int32 access);
+t_stat rom_reset (DEVICE *dptr);
+t_stat rom_boot (int32 u, DEVICE *dptr);
+t_stat rom_set_addr (UNIT *, int32, CONST char *, void *);
+t_stat rom_show_addr (FILE *, UNIT *, int32, CONST void *);
+t_stat rom_attach (UNIT *uptr, CONST char *cptr);
+t_stat rom_detach (UNIT *uptr);
+t_stat rom_help (FILE *st, DEVICE *dptr, UNIT *uptr, int32 flag, const char *cptr);
+t_stat rom_help_attach (FILE *st, DEVICE *dptr, UNIT *uptr, int32 flag, const char *cptr);
+const char *rom_description (DEVICE *dptr);
+
+/* ROM data structures
+
+   rom_dev       ROM device descriptor
+   rom_unit      ROM unit list
+*/
+
+#define unit_base u3
+#define unit_end u4
+
+#define ROM_UNITS 4
+#define ROM_UNIT_FLAGS UNIT_RO|UNIT_MUSTBUF|UNIT_BUFABLE|UNIT_ATTABLE
+
+DIB rom_dib[ROM_UNITS];
+
+MTAB rom_mod[] = {
+    { MTAB_XTD|MTAB_VUN|MTAB_VALR, 010, "ADDRESS", "ADDRESS",
+        &rom_set_addr, &rom_show_addr, NULL, "Bus address" },
+    { 0 }
+    };
+
+UNIT rom_unit[ROM_UNITS];
+
+DEVICE rom_dev = {
+    "ROM", rom_unit, NULL, rom_mod,
+    ROM_UNITS, 8, 9, 2, 8, 16,
+    rom_ex, rom_dep, rom_reset,
+    &rom_boot, &rom_attach, &rom_detach,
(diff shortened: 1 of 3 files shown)


larsbrinkhoff:
To run from the ROM, use something like this which is supposed to mimic a GT40.

```
set cpu 11/05
set cpu 16k
set dli enable
set dli address=17775610
at dli 3456
set vt enable
set vt crt=vr14
set vt scale=1
set vt address=17772000
set rom enabled
at rom bootvt.img
go 166000

**Comment by larsbrinkhoff** [comment]
Here is the ROM image file.  It's just a pure image.
[boot.zip](https://github.com/simh/simh/files/2348280/boot.zip)

**Comment by larsbrinkhoff** [comment]
@tlhackque suggested the ROM device should support multiple units.  That was my intention too, but I'd like advice on how to make multiple `DEVICE` instances.  Is there an example of this elsewhere in SIMH?

**Comment by larsbrinkhoff** [comment]
Here is another test ROM.

The code required the HITMISS register to be 1.  At the end, it gets an address from 704 and jumps to that plus 2.  I don't know how that's supposed to work.

```
set cpu 11/70
set cpu 64k
set rom enabled
set rom address=17765000
at rom pdp1170.img
dep hitmiss 1
go 165000
```
[1170.zip](https://github.com/simh/simh/files/2353613/1170.zip)

larsbrinkhoff:
Information on the SIMH mailing list is that the "1170" ROM is diagnostics, which is called from a boot ROM.  The boot ROM puts a return address, minus 2, in 704.

larsbrinkhoff:
I rebased this branch on top of the latest changes.  Verified that:

- The VT11 can read from the IO page ROM without my hack.
- Keyboard input from the VT window works with the configuration used to simulate a GT40.

Note: type `}R` to make the ROM do a restart and clear the screen.  I have yet to test `}L` to enter loading mode.

larsbrinkhoff:
I noticed pdp11_vt.c now fails to build when USE_DISPLAY is not set.  I added a commit here to fix that.

markpizz:
Opps.  Never mind.  I was looking in the wrong directory.  Sorry for any confusion.

larsbrinkhoff:
Admittedly, I do a lot of rebasing and rewriting to get a pull request into acceptable shape.  So I have been known to confuse people.  Glad you sorted it out.

markpizz:
Several details of feedback:

1) The code in rom_attach() should use a single call to sim_fread() to populate the rom_image buffer rather than the while loop and word assembly.  sim_fread() will do word assembly for you.  In fact, you can set unit flags UNIT_BUFFABLE|UNIT_MUSTBUF and call attach_unit() to do all the work.  The uptr->filebuf will then hold the ROM buffer's contents instead of you having to allocate it (rom_image goes away).  You might want to "sim_switches |= SWMASK ("Q");" to silence the attach message about buffering file in memory.  If you use attach_unit as I suggest here, you won't even need to call sim_fread().
2) In rom_make_dib() the vector info should be explicitly left as 0 since ROMs don't have interrupts, likewise for the interrupt ack and ulnt.  Since the number of units is a compile time value, you might just want to have a DIB array which has a dimension of ROM_UNITS rather than having to dynamically allocate the DIBs.
3) In rom_make_dib() you need to link the ROM device's DIBs together with the DIB next field in order for a device to have multiple dibs.  You should handle someone using ROM1 (unit 1) first and then ROM0 (unit 0), or if ROM_UNITS > 2, an arbitrary attach order.  The linked list doesn't have to be in attach order it, but all attached DIBs should be linked.  The beginning of the linked list should be stored in the rom_dev.ctxt cell.
4) rom_reset() should NOT free up anything.  rom_reset() will be called often, but for a ROM  device it probably shouldn't do anything.  Each unit should be freed (and its address space freed with build_ubus_tab(&rom_dev, NULL) re-connect any remaining DIBs and then if any remain call build_ubus_tab(&rom_dev, remaining-dib-listhead))  This freeing belongs in rom_detach().
5) rom_set_addr() should return SCPE_ALATT if the unit is already attached.  You can change an attached ROM address. 
6) for now, the range test on the ROM address requiring it be in the I/O page belongs there.  Once the memory plumbing is changed, then ROMs below the I/O page will be supported, but registering them will probably not be via build_ubus_tab().  You might want to use "return sim_messagef(SCPE_ARG. "ROM must be in I/O page, above 0%o\n", IOPAGEBASE);"

That is everthing I see at first glance.  :-)

larsbrinkhoff:
Thanks.  I don't know how to use the SIMH facilities, so it's good with some advice.

markpizz:
Further feedback:

1) UNIT_IDLE is not useful for a ROM device.  UNIT_IDLE tells the SCP event dispatcher that if the unit is at the top of the event queue, idling is possible.  Since no ROM unit will ever be on the event queue, it serves no purpose.  
2) All ROM units should have the same initial uptr->flags.  Setting these in rom_reset() with |= would be fine.  rom_reset() is called before the first simh command is read.  Appropriate initial flags should be: UNIT_RO|UNIT_MUSTBUF|UNIT_ATT|UNIT_BUFFABLE|UNIT_ATTABLE.  As I said previously, rom_reset() will be called any number of times (at simulator startup and during boot and whenever the CPU executes a RESET instruction).  uptr->unit_base and uptr->unit_end and uptr->flags&UNIT_ATT SHOULD NOT be cleared in rom_reset().  
3) rom_attach() shouldn't be calling scp_attach_unit(), but merely attach_unit() so you don't need to declare scp_attach_unit(),  You should probably be doing an existence check for the specified file before calling attach_unit()....Maybe a 0 return from sim_fsize_name() could help with that.  Even if the file exists, if it is 0 in size, that's not a candidate for a useful ROM image.
4) when I earlier said: "rom_set_addr() should return SCPE_ALATT if the unit is already attached." I meant to also say: You CAN'T change an attached ROM's address.
5) in rom_attach(), if rom_make_dib() fails, you should unwind anything that's already been done to the unit so that it ends up looking like it did prior to the call to rom_attach().  I would do this cleanup in rom_detach(), which actually should do something - see 7.
6) in rom_make_dib(), rom_dib[0] should be initialized like all others AND it should start with dib->rd as NULL.  I'm thinking that dib->wr should never be populated, but I've asked Bob Supnik about this detail. 
 build_ubus_tab() does need to be called for all units as they are attached.
7) rom_detach() should zero the related dib->rd and dib->wr values and then call build_ubus_tab() with that dib to deregister the block of I/O page memory that has been connected to this ROM. It should also call detach_unit() to get the file closed and the buffer deallocated.
8) The SHOW IOSPACE command will help you see if things look right.

larsbrinkhoff:
Thanks.  I'm still going though your first list.

larsbrinkhoff:
`SHOW IOSPACE` looks right, but `go 166000` no longer works for the VT11 ROM.

Do I need some kind of REG definitions for `examine rom` to work?

markpizz:
More feedback:
1) DIB->wr should always be NULL for ROM devices.  The CPU can't write to a ROM device.  Meanwhile, you could still support DEPOSIT to an attached ROM device if you wanted to.  You could interpret the address supplied in the DEPOSIT context in one of two ways: a) As an offset into the particular ROM unit's data, or b) as a full memory address, which you will have to explicitly range check AND mask and bound by the bus width (Q22 or U18).
2) I would have rom_reset() always uptr->flags |= ROM_UNIT_FLAGS; none of these ever get turned off.  This will then end up with the flags always set correctly when ROM_UNITS gets changed without having to manually add extra unit definitions with static flag assignments.
3) Where you have: uint16 *image = rom_unit[i].filebuf;, this should be uint16 *image = (uint16 *)rom_unit[i].filebuf; or we'll see compile warnings sometimes.

larsbrinkhoff:
I added in the changes you emailed me.  The VT11 ROM seems to work, but I'll have to test it more later.

I'm planning to squash all these commits in preparation for merging.

larsbrinkhoff:
Booting from the VT11 ROM seems to work.  I can no longer login to ITS, but I guess that's a problem elsewhere.

larsbrinkhoff:
Ah, it's just my CapsLock key is no longer mapped to Control.  Everything works OK with 122a87e.

PR #616, opened 2018-09-10
Attempting a Panda Display frontpanel

larsbrinkhoff:
This is a little premature for a pull request, but maybe I can get some input.

This first adds a CPU setting for enabling console lights.  Then an attempt at using this with a frontpanel application for the [Panda Display](https://gitlab.com/DavidGriffith/panda-display).  The frontpanel seems to start the simulator, but nothing happens.  The telnet console is unresponsive.

### DIFF
--- a/PDP10/pdp10_cpu.c
+++ b/PDP10/pdp10_cpu.c
@@ -204,6 +204,7 @@ int32 hst_p = 0;                                        /* history pointer */
 int32 hst_lnt = 0;                                      /* history length */
 InstHistory *hst = NULL;                                /* instruction history */
 int32 apr_serial = -1;                                  /* CPU Serial number */
+d10 console_lights = 0;
 
 /* Forward and external declarations */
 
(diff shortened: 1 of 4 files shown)


markpizz:
Relevant info to look into what you're doing here is the contents of the input files you're working with...

larsbrinkhoff:
Right, I pass in the path to the `pdp10` binary, and the configuration file which is this:

```
set console -u telnet=1927
set cpu its
set cpu idle
set cpu lights
set tim y2k
set dz 8b lines=8
at dz0 10004
set rp0 rp06
at rp0 out/simh/rp0.dsk
```

I can upload the rp0.dsk file somewhere if necessary.

markpizz:
If you zip the rp0.dsk file and the result isn't crazy big, then you can drop the zip file here, otherwise please put it on a google drive somewhere

larsbrinkhoff:
Here's the rp0.dsk image:  
http://lars.nocrew.org/its/its.zip

markpizz:
Looking at your panda.c, you have just gotten far enough to deliver the boot command to the simulator, but then your program exits.  

It doesn't continue to run getting call backed data deliveries and interacting with the simulator.  The example/test program interacts with the user until the user specifically wants to exit and then it cleanly shuts things down.  Your program needs to something similar.

markpizz:
While you get the basics put together here, you might want to stub out the USB interactions (and potentially emit text) to make the simulator interactions more robust...

larsbrinkhoff:
At some point I had it loop like this, but that didn't make a difference.

Anyway, I have come to feel unhappy about this approach.  The frontpanel API is a good fit for e.g. Blinkenbone or PiDP-8, but it's not what I had in mind for the Panda Display.

```
  while (sim_panel_get_state (panel) == Run) {
    usleep (100000);
  }
```

markpizz:
If you manually start the simulator, and enter the configuration commands you provided and then enter "BOOT RP0" which is the equivalent of what panda.c does, notice what is happening...

If you look at what the FrontPanelTest.c program is doing in its startup commands, on a Linux host you'll see: "! nohup xterm -e 'telnet localhost 1927' &\n"

The point here is that something needs to be connected to the console port of the simulated system.  An advanced configuration option (when your frontpanel app is more stable) would be to allow a console port connection to be arbitrarily delayed.  This would be done with: SET CONSOLE BUFFERED in the initial configuration file.

Once a console connection is in place (or deferred with CONSOLE BUFFERED), the BOOT command (sim_panel_exec_boot (panel, "RP0")) will be able to complete and the simulator will be in a RUN state.

PR #636, opened 2018-11-18
Unibus CH11: Chaosnet interface

larsbrinkhoff:
The CH11 is a Unibus device which can be used with PDP-11, VAX, and KS10.  It's a network interface for MIT's Chaosnet.

Software that should run on SIMH that supports this device include:

- ITS
- Berkeley Unix with MIT patches
- MINITS, a PDP-11 Chaosnet router/terminal concentrator

Outside SIMH, there's KLH10 and Lisp machine simulators.  Various encapsulating transport mechanisms exist: UDP, IP, Ethernet.

Documentation:  
https://lm-3.github.io/amber.html#Hardware-Programming-Documentation

### DIFF
--- a/PDP10/pdp10_defs.h
+++ b/PDP10/pdp10_defs.h
@@ -711,6 +711,7 @@ typedef struct pdp_dib DIB;
 #define INT_V_CR        27                              /* CD20 (CD11) */
 #define INT_V_DUPRX     28                              /* DUP11 */
 #define INT_V_DUPTX     29
+#define INT_V_CH        30                              /* CH11 Chaosnet */
 
 #define INT_RP          (1u << INT_V_RP)
 #define INT_TU          (1u << INT_V_TU)
@@ -728,6 +729,7 @@ typedef struct pdp_dib DIB;
 #define INT_CR          (1u << INT_V_CD)
 #define INT_DUPRX       (1u << INT_V_DUPRX)
 #define INT_DUPTX       (1u << INT_V_DUPTX)
+#define INT_CH          (1u << INT_V_CH)
 
 #define IPL_RP          6                               /* int levels */
 #define IPL_TU          6
--- a/PDP10/pdp10_ksio.c
+++ b/PDP10/pdp10_ksio.c
@@ -2009,6 +2009,8 @@ AUTO_CON auto_tab[] = {/*c  #v  am vm  fxa   fxv */
         {0000540}, {0540} },                             /* KMC11-A comm IOP-DUP ucode - fx CSR, fx VEC */
     { { "DMR" },         1,  2,  0, 0, 
         {0004000}, {0610} },                             /* DMR11 comm - fx CSR, fx VEC */
+    { { "CH" },          1,  1,  0, 0, 
+        {04140}, {0270} },                               /* CH11 - CHAOS Net - fx CSR, fx VEC */
 #else
     { { "QBA" },         1,  0,  0, 0, 
         {017500} },                                     /* doorbell - fx CSR, no VEC */
--- a/PDP10/pdp10_sys.c
+++ b/PDP10/pdp10_sys.c
@@ -62,6 +62,7 @@ extern DEVICE xu_dev;
 extern DEVICE dup_dev;
 extern DEVICE kmc_dev;
 extern DEVICE dmc_dev;
+extern DEVICE ch_dev;
 extern REG cpu_reg[];
 extern a10 saved_PC;
 
@@ -99,6 +100,7 @@ DEVICE *sim_devices[] = {
     &dup_dev,
     &kmc_dev,
     &dmc_dev,
+    &ch_dev,
     NULL
     };
 
--- a/PDP11/pdp11_ch.c
+++ b/PDP11/pdp11_ch.c
@@ -1,0 +1,537 @@
+/* pdp11_ch.c: CH11 Chaosnet interface.
+  ------------------------------------------------------------------------------
+
+   Copyright (c) 2018, Lars Brinkhoff.
+
+   Permission is hereby granted, free of charge, to any person obtaining a
+   copy of this software and associated documentation files (the "Software"),
+   to deal in the Software without restriction, including without limitation
+   the rights to use, copy, modify, merge, publish, distribute, sublicense,
+   and/or sell copies of the Software, and to permit persons to whom the
+   Software is furnished to do so, subject to the following conditions:
+
+   The above copyright notice and this permission notice shall be included in
+   all copies or substantial portions of the Software.
+
+   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+   IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+   FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+   THE AUTHOR BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+   IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+   CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+   Except as contained in this notice, the name of the author shall not be
+   used in advertising or otherwise to promote the sale, use or other dealings
+   in this Software without prior written authorization from the author.
+
+  ------------------------------------------------------------------------------
+
+*/
+
+#if defined (VM_PDP11)
+#include "pdp11_defs.h"
+
+#elif defined (VM_PDP10)
+#include "pdp10_defs.h"
+
+#elif defined (VM_VAX)
+#include "vax_defs.h"
+
+#else
+#error The CH11 device only works with Unibus machines.
+#endif
+
+#include "sim_tmxr.h"
+
+/* CSR bits */
+#define TIE   0000001 /* Timer interrupt enable */
+#define LOOP  0000002 /* Loop back */
+#define SPY   0000004 /* Loop back */
+#define CRX   0000010 /* Clear receiver */
+#define RXIE  0000020 /* Receive interrupt enable */
+#define TXIE  0000040 /* Transmit interrupt enable */
+#define TXA   0000100 /* Transmit abort */
+#define TXD   0000200 /* Transmit done */
+#define CTX   0000400 /* Clear transmitter */
+#define LOST  0017000 /* Lost count */
+#define RESET 0020000 /* Reset */
+#define CRC   0040000 /* CRC error */
+#define RXD   0100000 /* Receive done */
+
+#define STATUS_BITS (TIE|LOOP|SPY|RXIE|TXIE|TXA|TXD|LOST|CRC|RXD)
+#define COMMAND_BITS (TIE|LOOP|SPY|RXIE|TXIE)
+
+BITFIELD ch_csr_bits[] = {
+  BIT(TIE),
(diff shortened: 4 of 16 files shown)


larsbrinkhoff:
This is now tested working with vax780 running 4.1BSD, talking to a PDP-10 running ITS.

I converted the issue to a pull request.

markpizz:
Several points of feedback:
1) pdp11_ch.c should have CRLF line endings to be consistent with the rest of the simh codebase.
2) the bit specified in vax860_defs.h for INT_V_CH is wrong.
3) in pdp11_ch.c, what is the thinking behind the conditional MTAB entries specific for VM_PDP11?
4) The CH DEVICE structure doesn't populate the REG member with a pointer to simh REG declarations which describe the state of the device.  All variables used by the device should be included.  These will then be visible with SCP EXAMINE and DEPOSIT commands.
5) You've got ch_ex and ch_dep routines which don't do anything.  They probably shouldn't be there at all.
6) There is no reason that the tmxr transport for this data has to be UDP.  It should work correctly independent of UDP or TCP.  Let the user specify their desired transport (AND incoming port number) on the SCP ATTACH command.  The current hard setting of the incoming port number to 42043 won't work on all platforms since if both ends of the conversation are on the same host system, there will be two processes listening on the same port.  The attach string that is passed into tmxr_open_master() should have "PACKET" indicated along with the other relevant values.  
7) You probably want an MTAB entry to specify the address of the peer simulator (see pdp11_dmc.c for an example of that and it's attach command).
8) The ch_svc unit service routine for received packets schedules itself to run every 1000 instructions.  This will effectively block a simulator's ability to idle.  Things should work just fine if you used sim_clock_coshedule().
9) When you make the change suggested in 6 above, you should move the call to tmxr_poll_conn out of the ch_attach() routine and perform that activity in ch_svc().  ch_svc should be activated in the ch_attach() routine after a successful call to tmxr_open_master() and canceled in the ch_detach() routine.

larsbrinkhoff:
Thanks.  I'll go through your list.  For now:

3. This is just copying and pasting from elsewhere.
6. The reason for using UDP on port 42042 is to interoperate with another existing implementation.  I just updated the pull request so maybe you saw a previous revision, but now there's a PORT device configuration.

markpizz:
I guess that I did see an earlier version since that would explain the odd branch structure I saw after I had picked up your stuff.


> 3. This is just copying and pasting from elsewhere. Please advise.

Well, there may be some discussion about why that stuff is in the other place(s) you may have seen it, but for your device it only ever had a single address and vector, so the non-pdp11 code is the right stuff.

> 6. The reason for using UDP on port 42042 is to interoperate with another existing implementation. I just updated the pull request so maybe you saw a previous revision, but now there's a PORT device configuration.

If you remove the hard coded UDP in the attach string, someone who wanted to interoperate with the other implementation could merely specify it as part of the attach string:  "localhost:42042,UDP" and that would work compatibly.  I guess you could leave the UDP there, and someone who wanted TCP could specify the attach string: "localhost:42042,TCP".  In any case, the tmxr_poll_conn() belongs in the ch_svc() routine and the ch_unit should be scheduled after a successful call to tmxr_open_master() and it should be sim_cancel(&ch_unit) in the ch_detach() routine.  Populate the ch_help() routine to explain the usage or more specifically provide a help_attach() routine which explains the usage and the ch_help() can describe things mentioned in https://lm-3.github.io/amber.html#Hardware-Programming-Documentation

ForthStar:
I'm addressing your points.  For some reason number 9 broke the device, but I'll have to check on that tomorrow.  The help text is still to do.

larsbrinkhoff:
I'm addressing your points. For some reason number 9 broke the device, but I'll have to check on that tomorrow. The help text is still to do.

larsbrinkhoff:
I believe these additional commits adresses all your points.  So I'll be awaiting further review comments.

When done, I intend to squash my commits before merging.  Do you want your commits squashed or to remain as is?  Or maybe you want to handle these details.  Let me know.

markpizz:
Squashing all commits is preferred. 

Further comments:

1) The DEVICE reset routine is part of the SCP plumbing and generally not a routine that would be called by any device code due to programmatic behavior of code within a simulator.  It's job is to perform initialization and setup interactions with the rest of the SCP framework.  It may also call routines within the device simulation that reset state.  I mention this since currently ch_command() calls ch_reset().  Maybe there should be another routine called ch_prog_reset() which ch_command() and ch_reset() call...
2) The info you've got in the help routine really belongs in the separate REG and MTAB entries.  For example in the PDP11 simulator, type HELP DZ REG and HELP DZ SET and HELP DZ SHOW and EXAMINE DZ STATE.  The REG and MTAB arrays are walked by the SCP HELP logic to gather that output.  It is preferred that new devices use the hierarchical help model (see HELP DMC or HELP KDP), however it is also acceptable to use the simpler non-hierarchical help model that  is just one blob of output, but uses the related device help routines (see HELP XU).
3) You might want to define a BITFIELD structure that describes the details of the CH device's CSR which you can then display in debug output (see DZ debug in dz_wr() and dz_rd() as a examples).
4) One of the purposes of the DEVICE REG array is to contain ALL of the device's state information so that a simulator can be stopped with ^E a SCP "SAVE simu.save" command executed, and the simulator exits, and the simulator is started again and a SCP "RESTORE simu.save" command executed and a "CONTINUE" command executes and absolutely nothing in the simulator can tell that things weren't running the whole time.  Any configuration data your device gathers through SCP SET commands is generally part of that state (the SET CH NODE=value for example).  This information usually would be described in REG structures with REG_HRO flag bits set so that EXAMINE and DEPOSIT commands don't display or change these things.
5) One change I made to ch_svc() really shouldn't be there.  Specifically the setting of ch_lines[0].rcve to TRUE on a successful return from tmxr_poll_conn().  The rcve state is properly managed by your existing register interaction logic.

markpizz:
More comments:
6) There needs to be some robust validation logic in the SET CH PEER.  Copy from dmc_setpeer().
7) After comment 4 above is addressed, to be properly behaved during a RESTORE command, the last few lines of ch_attach() should be:
```
  uptr->filename = (char *)realloc (uptr->filename, 1 + strlen (cptr));
  strcpy (uptr->filename, cptr);
  sim_clock_coschedule (uptr, 1000);
  return SCPE_OK;
```

ForthStar:
Ok, round two finished.

ForthStar:
The help text now looks like this:

```
CH11 Chaosnet interface

The CH11 is a Unibus device which can be used with PDP-11, VAX, and
KS10.  It's a network interface for MIT's Chaosnet.  Options allow
control of the node address and network peer.  The node address must
be a 16-bit octal number.

CH device SET commands:

set CH ADDRESS=val              Unibus address
set CH VECTOR=val               Interrupt vector
set CH PEER=val                 Remote host name and port
set CH NODE=val                 Chaosnet node address
set CH ENABLE                   Enables device CH
set CH DISABLE                  Disables device CH
set CH DEBUG                    Enables debugging for device CH
set CH NODEBUG                  Disables debugging for device CH
set CH DEBUG=TRC;REG;PKT;DAT;INT;ERR
                                Enables specific debugging for device CH
set CH NODEBUG=TRC;REG;PKT;DAT;INT;ERR
                                Disables specific debugging for device CH

*CH device DEBUG settings:
    TRC         Detailed trace
    REG         Hardware registers
    PKT         Packets
    DAT         Packet data
    INT         Interrupts
    ERR         Error conditions

Configured options and controller state can be displayed with:

CH device SHOW commands:

show CH ADDRESS                 Unibus address
show CH VECTOR                  Interrupt vector
show CH PEER                    Remote host name and port
show CH NODE                    Chaosnet node address
show CH DEBUG                   Displays debugging status for device CH

The CH11 simulation will encapsulate Chaosnet packets in UDP or TCP.
To access the network, the simulated Chaosnet interface must be attached
to a network peer.

To configure CH11, first set the local Chaosnet node address, and
the peer:

  sim> SET CH NODE=<octal address>
  sim> SET CH PEER=<remote host>:<remote port>

Then, attach a local port.  By default UDP is used:

  sim> ATTACH CH <local port>

If TCP is desired, add "TCP":

  sim> ATTACH CH <local port>,TCP

Software that runs on SIMH that supports this device include:
 - ITS, the PDP-10 Incompatible Timesharing System
 - Berkeley Unix with MIT patches
 - MINITS, a PDP-11 Chaosnet router/terminal concentrator

Outside SIMH, there's KLH10 and Lisp machine simulators.  Various
encapsulating transport mechanisms exist: UDP, IP, Ethernet.

Documentation:
https://lm-3.github.io/amber.html#Hardware-Programming-Documentation

**Comment by larsbrinkhoff** [comment]
The help text now looks like this:

```
CH11 Chaosnet interface

The CH11 is a Unibus device which can be used with PDP-11, VAX, and
KS10.  It's a network interface for MIT's Chaosnet.  Options allow
control of the node address and network peer.  The node address must
be a 16-bit octal number.

CH device SET commands:

set CH ADDRESS=val              Unibus address
set CH VECTOR=val               Interrupt vector
set CH PEER=val                 Remote host name and port
set CH NODE=val                 Chaosnet node address
set CH ENABLE                   Enables device CH
set CH DISABLE                  Disables device CH
set CH DEBUG                    Enables debugging for device CH
set CH NODEBUG                  Disables debugging for device CH
set CH DEBUG=TRC;REG;PKT;DAT;INT;ERR
                                Enables specific debugging for device CH
set CH NODEBUG=TRC;REG;PKT;DAT;INT;ERR
                                Disables specific debugging for device CH

*CH device DEBUG settings:
    TRC         Detailed trace
    REG         Hardware registers
    PKT         Packets
    DAT         Packet data
    INT         Interrupts
    ERR         Error conditions

Configured options and controller state can be displayed with:

CH device SHOW commands:

show CH ADDRESS                 Unibus address
show CH VECTOR                  Interrupt vector
show CH PEER                    Remote host name and port
show CH NODE                    Chaosnet node address
show CH DEBUG                   Displays debugging status for device CH

The CH11 simulation will encapsulate Chaosnet packets in UDP or TCP.
To access the network, the simulated Chaosnet interface must be attached
to a network peer.

To configure CH11, first set the local Chaosnet node address, and
the peer:

  sim> SET CH NODE=<octal address>
  sim> SET CH PEER=<remote host>:<remote port>

Then, attach a local port.  By default UDP is used:

  sim> ATTACH CH <local port>

If TCP is desired, add "TCP":

  sim> ATTACH CH <local port>,TCP

Software that runs on SIMH that supports this device include:
 - ITS, the PDP-10 Incompatible Timesharing System
 - Berkeley Unix with MIT patches
 - MINITS, a PDP-11 Chaosnet router/terminal concentrator

Outside SIMH, there's KLH10 and Lisp machine simulators.  Various
encapsulating transport mechanisms exist: UDP, IP, Ethernet.

Documentation:
https://lm-3.github.io/amber.html#Hardware-Programming-Documentation

larsbrinkhoff:
Squashed.  Checked that VAX is still working.

markpizz:
Why does the squashed commit have a date some 16 days ago rather than just now?

larsbrinkhoff:
I updated the date using the commit --date option.

markpizz:
Date looks good, but I'm getting some compiler warnings about signed/unsigned inconsistent types on Windows.

in ch_set_node(), please change "x = get_uint (cptr, 8, 0177777, &r);" to "x = (int)get_uint (cptr, 8, 0177777, &r);".

larsbrinkhoff:
Fixed.

larsbrinkhoff:
Related question:  
What's the difference in SIMH, between a Unibus device and a Qbus one?  Apparently the MINITS Chaosnet software for PDP-11 ran on both Unibus and Qbus machines, and there was a Qbus Chaosnet interface.

larsbrinkhoff:
FYI, there is a plan to use [MINITS](http://gunkies.org/wiki/MINITS) running on SIMH to attach Multics to Chaosnet.

PR #639, opened 2018-12-09
BESM6: Printer file position must be updated.

leobru:
Why would a *sequential* file need to be re-seeked in scp.c  to the stored position  rather than to the end of file at every restart of the simulation from the CLI is unclear, but a 1-line fix solves the issue.

### DIFF
--- a/BESM6/besm6_printer.c
+++ b/BESM6/besm6_printer.c
@@ -144,6 +144,7 @@ void printer_control (int num, uint32 cmd)
     case 1:         /* linefeed */
         READY &= ~(PRN1_LINEFEED >> num);
         offset_gost_write (num, u->fileref);
+        u->pos = ftell(u->fileref);
         dev->feed = LINEFEED_SYNC;
         break;
     case 4:         /* start */


markpizz:
> Why would a sequential file need to be re-seeked in scp.c to the stored position rather than to the end of file at every restart of the simulation from the CLI is unclear, but a 1-line fix solves the issue.

Well, the pos field of the unit structure (for sequential devices - UNIT_SEQ) is commonly mentioned in an explicit REG declaration (often named POS) which then enables the user to "reposition" a sequential device through SCP (e.g., rewind a paper tape with "DEPOSIT POS 0"). 

This was the original purpose of maintaining the unit pos variable, so to directly support that activity, SCP repositions UNIT_SEQ devices to the indicated position whenever execution starts (RUN/GO/BOOT/STEP, etc.).

A secondary reason exists to assure that all UNIT_SEQ units are properly positioned across a SAVE/RESTORE command.  Previously, this proper positioning required an explicit REG declaration that included the unit pos variable.  However, while gathering info to answer your question, I determined that the unit pos variable should always be stored and restored in any SAVE/RESTORE.  This has been added to scp.c and the behaviour you encountered documented in doc/simh.doc.

Independent of this, while that thinking was happening other commits were submitted and merged.  So this change was merged with an explicit Cherry Pick commit attributing @leobru as the author.

To get a proper view of the simh/simh repo master branch locally and in your leobru/simh repo you should probably do:

    $ git reset --hard
    $ git push -f https://github.com/leobru/simh
    $ git pull https://github.com/simh/simh

PR #640, opened 2018-12-11
Fix A and M in Type 340 character set

larsbrinkhoff:
There are minor deviations in the Type 340 character set compared to the "340 DISPLAY PROGRAMMING MANUAL".  As far as I can see, only A and M are affected.  The compete sets are reproduced here for reference.

340sim
340doc

CC @philbudne

PR #641, opened 2018-12-11
fix for https://github.com/simh/simh/issues/640

philbudne:


markpizz:
Merged with Cherry pick to rewrite commit message in standard form.  Commit was done with Phil as the author.

PR #643, opened 2018-12-14
[3b2] Properly initialize DMAC Current Word Count

sethm:
The DMAC controller "Current Word Count" expects to decrement beyond 0 to 0xffff. Its initial value should also be 0xffff. Setting the initial value to 0 introduced a very subtle bug related to DUART self-test following a soft power reset.


sethm:
I just realized my commit message should not use brackets, but rather a colon to separate "3b2" and the summary. I'm going to close this pull request, rewrite my history, and then open a new one with the proper format.

PR #642, opened 2018-12-21
PDP11: NG - Tom Knight vector display

larsbrinkhoff:
Unibus device to control eight vector displays.

Documentation: Appendix A of AI Lab memo 240: https://dspace.mit.edu/bitstream/handle/1721.1/6195/AIM-240a.pdf

Applications:
- PDP-11 Logo
- Dazzle Dart


larsbrinkhoff:
@markpizz, @philbudne,

I invite comments on this.

larsbrinkhoff:
I don't seem to get any PCLK interrupts.

The code writes 115 (octal) to the PCLK CSR register, and it responds:

```
PCLK REG: pclk_wr(PA=0x003FF560 [CSR ], access=2, data=0x4D) ERR0 DONE0 IE1 FIX0 UPDN0 MODE1 RATE=line GO1
DBG(7)> PCLK SCHED: pclk_set_ctr(val=0) - delay=65536, rv=2, xtim[rv]=16667, usecs=1092288512
```

I checked in a debugger that SIMH does get here in pdp11_pclk.c pclk_set_ctr():

    sim_activate_after (&pclk_unit, usec_delay);

And usec_delay is 1092288512 as printed above.  That should be 1,092 milliseconds right?  Yet my breakpoint at pclk_svc doesn't trigger.

larsbrinkhoff:
No, of course it's 1092 seconds.

markpizz:
Well, 1092 seconds sounds like a long time.  This amount of time is derived from the CSR RATE bits which seem to be selecting the the line frequency (which defaults to 60Hz) as the counter units.  Do you think the KW11P should behave differently?

markpizz:
Note that the KW11P manual has a programming example:
image
This example uses a single mode vs the repeat mode specified when the CSR is set to 115, but we're still using the same 60Hz tick size, so with a count value initialize to 0 by the system reset, the overflow will happen after 65536/60 seconds and that is 1092...

larsbrinkhoff:
Thanks.  Then it looks like the application code is in error.  Maybe Dazzle Dart was loaded by something else, and that other code set CSB to 60.  I suppose updating the Dazzle Dart code is the easiest thing to do.

larsbrinkhoff:
To 1, I mean.  With that, the 30-second timer is counted down as expected.

larsbrinkhoff:
Maybe these buttons were used for input?

larsbrinkhoff:
To clarify, only the first two commits are for consideration for merging to the master branch.  The (currently) two others are for testing: programs using the display and inputs, and work in progress to use the SDL joystick API.

markpizz:
Sorry,  Due to holiday traveling and family activities, I really haven't had time to look closely at this.

Now that I look at where you've been going,  I see you're making direct SDL calls from simulator code in pdp11_box.c.  Maybe you do this while you figure out what generically needs to be done, but then a generic solution extends what sim_video.c provides (which is the simulator interface to SDL).

A completely different angle to look at this user interface/input concept might be to write a GUI application (which uses SDL or anything else to present the UI and gather its input) and then pass button presses, etc into the simulator using the frontpanel API.

larsbrinkhoff:
No problem, there's no rush.

I can see it was a bad idea to have experimental commits on the branch, so I removed them.  The direct calls to SDL should be gone now.  I'll rework that and submit it in another pull request.

So what we have here now is just the NG display device, and the input device.  The inputs are mapped from keyboard keys.  This isn't usable for playing the game, but good for testing.

larsbrinkhoff:
Made a tweak.  Now the "forbidden zones" are semi-circular.

daz5

larsbrinkhoff:
Obviously the name "box" is super bad.

I asked around, and apparently it was something custom made for this game.  So maybe "daz"?

larsbrinkhoff:
Ping @markpizz, do I need to fix anything here?

markpizz:
Several things:
1) daz_reset should ONLY install the KB event processor IF the device is enabled!
2) only one device should be able to have a vid_open device enabled at the same time.  Any other SET dev ENABLE should be rejected.  There are 3 devices which now can open a video display.
3) CH and NG devices seem to have the same bus address and vector.  You should not be able to enable both at the same time.  Look closely in pdp11_td.c and pdp11_dl.c for devices which must co-exist or be excluded simultaneously.
4) It is not obvious to me how these devices work.  Specifically if they do DMA, but there should be something which only enables things if the current CPU bus type is reasonable.
5) If either these devices exist with a single primary program that demonstrates their use, then that program should be included and probably loaded directly via a BOOT <dev>.  See BOOT VT for a similar example.
6) ng.c has two functions which must return a value:
7) USE_DISPLAY might be defined, but vid_open might return a failure status.  If so, then any attempt to SET <dev> ENABLE should leave the device disabled (DEV_DIS).

markpizz:
ng.c changes:
```
diff --git a/display/ng.c b/display/ng.c
index b270ca5c..ac7f71ac 100644
--- a/display/ng.c
+++ b/display/ng.c
@@ -117,12 +117,12 @@ ng_init(void *dev, int debug)
 
 static int fetch (int a, uint16 *x)
 {
-  ng_fetch (a + reloc, x);
+  return ng_fetch (a + reloc, x);
 }
 
 static int store (int a, uint16 x)
 {
-  ng_store (a + reloc, x);
+  return ng_store (a + reloc, x);
 }
 
 static void point (void)
```

markpizz:
Please try to run the PDP11 simulator when things are compiled with USE_DISPLAY:
```
mark@Ubuntu-18:~mark/simh-larsbrinkhoff$ BIN/pdp11
CH device in conflict with NG.
Fatal simulator initialization error
Unit already attached
Hit Return to exit: 
```

larsbrinkhoff:
> Fatal simulator initialization error

Yes, I had this problem, but I updated the branch to fix it.  It should be gone now.

markpizz:
Should enabling the DAZ device automatically require (or force) the NG device to be enabled and vice versa?

What about my point number 5 above?

larsbrinkhoff:
I haven't addressed point 5 yet.  Sometimes I have to sleep and work too. ;-)

The NG device should work without DAZ, like when running 11LOGO.  DAZ only have one application: the Dazzle Dart game.  So it could require NG.

larsbrinkhoff:
Regarding 5.  The primary NG application would be 11LOGO.  And the *only* DAZ application is Dazzle Dart.  So I can make BOOT start those.  (I do consider it pecuilar to bundle the target software inside the simulator, but hey it doesn't hurt.)

markpizz:
The goal of bundling these things is to make it very easy for a user to get started and see something without having to climb a configuration learning curve.  Otherwise the BOOT operation on these devices doesn't do anything.  Someone like this might then be inspired to dig deeper first for more games and subsequently other endeavors.

So, SET DAZ ENABLE should also enable NG.  BOOT DAZ should configure things to run Dazzle Dart and start it running.  BOOT NG should run 11LOGO...

sim_BuildROMs.c should have a line which produces the include file containing the Dazzle Dart program from the .lda file which belongs in PDP11/DazzleDart/...

larsbrinkhoff:
Thanks.  I'll see to it in the coming days.

larsbrinkhoff:
I added 11LOGO and Dazzle Dart.  BOOT NG and BOOT DAZ will run them.

markpizz:
I'll report details as I find them:
1:
start simulator:
```
    sim> set daz enable
    sim> set vt enable
    sim> boot daz
```
Please fix.
A variation on the same problem:
2:
start simulator:
```
    sim> set vt enable
    sim> set daz enable
    sim> boot vt
```
Please fix.
3:
start simulator:
```
    sim> set daz enable
    sim> boot daz
set focus to NG window
type a space (or any number of unexpected characters)... daz_keyboard() needs to be more accepting. :-)
```
Please fix.

markpizz:
4:
start simulator:
```
PDP-11 simulator V4.0-0 Current        git commit id: 390656b0
sim> set ng ena
NG device not compatible with system bus
sim> sh ng
NG      disabled
sim>
```
Problem:  NG video window appears anyway
Please fix.

larsbrinkhoff:
I did some rebasing and squashing.

Thanks for the feedback.  I'll make the requested changes later this week.

larsbrinkhoff:
> 1:
> start simulator:
> 
>     sim> set daz enable
>     sim> set vt enable
>     sim> boot daz
> 
> Please fix.
> A variation on the same problem:
> 2:
> start simulator:
> 
>     sim> set vt enable
>     sim> set daz enable
>     sim> boot vt
> Please fix.

I tried to fix these two, but I don't know how.  Please advice.

markpizz:
Pick up the changes at https://github.com/markpizz/simh in the lars/ngdis branch.

markpizz:
I didn't look closely enough before, but after trying the BOOT NG steps to invoke 11LOGO, it seems that a potential user absolutely needs more built in help.  Specifically what are the commands that the 11LOGO program accepts?  These should either be visible in the HELP NG output OR, better yet should be displayed when the program is booted.  Maybe there could be a series of EXPECT/SEND commands that demonstrate a few things.  I don't know.

larsbrinkhoff:
This version of 11LOGO isn't finished, so there is no help text and there are some bugs.  But it's the only version we have.  It's not very user friendly.

With that in mind, these are some commands to get started:

```
STARTDISPLAY - Start the display and enable turtle graphics.
FORWARD n - Move turtle n steps forward.
RIGHT n - Turn turtle n degrees to the right.
LEFT n - Turn turtle n degrees to the left.
```

Would this be appropriate for the help text?

There is also AI Lab memo 315, "A glossary of Logo primitives": https://dspace.mit.edu/bitstream/handle/1721.1/6228/AIM-315.pdf  
This may or may not match the software that remains today.

markpizz:
The AI Labs memo seems to be describing a much more advanced version of 11 Logo.  Too bad that is no longer available.

How about listing all of the currently valid commands when the NG device is booted.  This can be done with sim_printf() statements in the ng_boot() routine.  Without that info, how would someone playing with this have a clue about what commands might work?

Digging through the source in the PDP11/11logo directory, I find that several of the files here have inconsistent line endings.  Since these are text files, all line endings should be CRLF.

I notice in the file system.327 there is a list of commands.  This is significantly long.  Maybe the boot should display this list along with a pointer to the AI labs memo 315 with the caveat that not everything may be the same.

markpizz:
I rewrote history at https://github.com/markpizz/simh in the lars/ngdis branch and committed the line-endings change mentioned in the last comment.

larsbrinkhoff:
Brought in your changes, and added this text to `boot ng`:

```
List of 11LOGO commands:
AND, BACK, BUTFIRST, BUTLAST, COUNT, CTF, DIFFERENCE, DISPLAY, DO,
EDIT, ELSE, EMPTYP, END, EQUAL, ERASETRACE, FIRST, FORWARD, FPRINT,
FPUT, GO, GREATER, HEADING, HERE, HIDETURTLE, HOME, IF, KILLDISPLAY,
LAMPOFF, LAMPON, LAST, LEFT, LESS, LEVEL, LIST, LISTP, LPUT, MAKE,
MOD, NEWSNAP, NUMBERP, OF, OUTPUT, PENDOWN, PENUP, PRINT, PRODUCT,
QUOTIENT, REQUEST, RIGHT, RUG, SENTENCE, SETHEADING, SETTURTLE, SETX,
SETXY, SETY, SHOW, SHOWTURTLE, SNAP, STARTDISPLAY, STF, STOP, SUM,
THEN, TO, TOOT, TRACE, TYPE, VERSION, WIPE, WIPECLEAN, WORD, WORDP,
XCOR, YCOR.

MIT AI memo 315 documents a later version of 11LOGO but may be helpful
in exploring the software.  It can currently be found here:
https://dspace.mit.edu/handle/1721.1/6228

To get started with turtle graphics, type STARTDISPLAY.

markpizz:
That looks good.

Feel free to squash my changes away and maybe update the date on these commits to today.

larsbrinkhoff:
Ok, all freshened up.

PR #648, opened 2018-12-29
H316 plt

apwise:
This adds a plotter device to the H316 simulator. This was an option supplied and supported by Honeywell, though it is certainly one of the lesser know options.
Beyond the additions to the simh simulator, there's a separate C++ program (h16-plt2ps) to convert the output into PostScript so it can be viewed, and a small example H316 program (tree.f compiled and linked as tree.slst).


larsbrinkhoff:
I'm also interested in adding plotter support.  To the KA10 emulator.

PR #652, opened 2019-01-07
display: Update Type 340 character rendering

larsbrinkhoff:
The Type 340 display (or really the Type 342 character generator) has two character sets (shift in and shift out) and shift characters to choose between them.  The first set is documented in the PDP-7 "340 DISPLAY PROGRAMMING MANUAL", but the other set is not.

@philbudne inferred the existence of lower case characters from PDP-6 Spacewar.  I have now found a translation table in ITS, from ASCII codes to 340 codes.

| ASCII Character | Type 340 Code | Comment
| --- | --- | ---
| ^H | 72 | Backspace?
| ^G | 63 | Unclear what glyph
| ^[ | 47 | Right arrow
| | 50 | Down arrow?
| [ | 53
| \\ | 52
| ] | 54
| ^ | 67 | Non spacing character
| _ | 60
| ` | 66 | Non spacing character
| { | 55
| \| | 62
| } | 56
| ~ | 43


larsbrinkhoff:
CC @philbudne.

display/type340.c
```diff
@@ -670,27 +674,27 @@ static const unsigned char chars[128][5] = {
...
+    { 0000, 0000, 0202, 0202, 0376 },   /* 54 ] */
+    { 0020, 0154, 0202, 0000, 0000 },   /* 55 { */
+    { 0000, 0000, 0202, 0154, 0020 },   /* 56 } */
     { 0004, 0010, 0020, 0040, 0100 },   /* 57 / */
-    { 0174, 0212, 0222, 0242, 0174 },   /* 60 0 */
+    { 0002, 0002, 0002, 0002, 0002 },   /* 60 _ */
     { 0000, 0102, 0376, 0002, 0000 },   /* 61 1 */
-    { 0116, 0222, 0222, 0222, 0142 },   /* 62 2 */
+    { 0000, 0000, 0376, 0000, 0000 },   /* 62 | */
     { 0104, 0202, 0222, 0222, 0154 },   /* 63 3 */
```
philbudne: I think we should just remove all the duplicates from the "shifted" character set.  They were only there as place-holders.

Maybe replace them with 0376,  0376,  0376,  0376,  0376,  0376 (black square) to indicate that it's a (black) hole in our knowledge?

Did you add a circumflex?  Given that there are right and down arrows, and the age of the hardware, up-arrow might be more likely....

display/type340.c
```diff
@@ -701,6 +705,14 @@ static const unsigned char chars[128][5] = {
     { 0100, 0200, 0236, 0220, 0140 }    /* 77 ? */
 };
 
+void
+ty340_set_grid(int w, int h)
```
philbudne: type342_set_grid?

display/type340.c
```diff
@@ -732,6 +744,11 @@ character(int n, unsigned char c)
     case CHRESC:                        /* escape */
         return 1;
     }
+    /* account for non spacing characters */
```
philbudne: Here I think we should replace the case stmt and the new conditions below with flags stored in a 6th byte in the character table.  It would be cleaner, and likely more like how the actual hardware did things.....

Yeah, there are five available low order bits, but it would be neater to have all the flags in one place, unless there need to be more than 8!!!

display/type340.c
philbudne: I was thinking of the original ASCII, which had up-arrow.....  The PDP-6 was certainly at the cusp between six-bit and ASCII, I forget where in the 18-bit lineage the change occurred (maybe between the PDP-4 and PDP-7?)

larsbrinkhoff: Thanks, I'll replace the unknown characters.

Yes, I added circumflex.  Good point; it's not unlikely there was an up arrow instead.  But if so, I think the underscore should be a left arrow.

larsbrinkhoff:
This is the appearance of the printable ASCII characters as translated by ITS' DIS device:
340chars

My contribution are the characters [ \\ ] ↑ ← ` { | } ~

Not visible here are down arrow and right arrow.

larsbrinkhoff:
The commits marked [REVIEW FIX] will be squashed back when the pull request is in an acceptable state.

display/type340.c
```diff
@@ -711,37 +731,47 @@ character(int n, unsigned char c)
...
+    }
+    if (flags & CH_LC) {                /* "SHIFT OUT (VERT)" */
         u->shift = 0100;                /* lower case in SPCWAR 163 */
         return 0;
-    case CHRESC:                        /* escape */
+    }
+    if (flags & CH_ESC) {               /* escape */
         return 1;
     }
+    if ((flags & CH_NSPC) && u->xpos >= u->width*s) {
```
philbudne: Seems ugly (to me) to subtract after the fact; wrap the increment in `!(flags &CH_NSP)` instead?

larsbrinkhoff: It seems the nonspacing characters backspace and overstrike the character before.  This decrement is not after the fact, but before the character is put on the display.  The increment happens after.

larsbrinkhoff:
Updated default to 6x9.  Verified building and running with KA10 simulator.

I'm squashing the fix commits to make a clean pull request.

markpizz:
These changes have been merged into the master branch, but this case is strange since no current code in the master branch actually uses the affected code.

larsbrinkhoff:
Thanks.  Yes, it's a strange situation since some code lives in the KA10 simulator, and some here.

@rcornwell, will you please bring in these commits at a time convenient to you?

PR #653, opened 2019-01-11
display: Update 340 simulation to use sim_ws.

larsbrinkhoff:
I'm not sure if this is appropriate here.  It kind of straddles upstream SIMH (this repo) and @rcornwell's KA10 simulator.

The background is that I want the Type 340 simulation to use sim_ws.c instead of x11.c.  This is because I want the new `vid_display_kb_event_process` API.

To do this, I need to pass the SIMH dptr from the KA10 simulator, to `display_init` which will in turn pass it on to `ws_init`.  The sim_ws `ws_init` expects the dptr instead of the UNIT * passed to the x11 `ws_init`.

There's a corresponding change to the top-level makefile which replaces x11.c with sim_ws.c, but it doesn't make sense for this repository.  There's also a change to ka10_dpy.c of course.

### DIFF
--- a/display/type340.c
+++ b/display/type340.c
@@ -144,12 +144,12 @@ ty340_set_dac(ty340word addr)
 #endif
 
 ty340word
-ty340_reset(void)
+ty340_reset(void *dptr)
 {
     struct type340 *u = UNIT(0);
 #ifndef TY340_NODISPLAY
     if (!u->initialized) {
-        display_init(DIS_TYPE340, 1, u); /* XXX check return? */
+        display_init(DIS_TYPE340, 1, dptr); /* XXX check return? */
         u->initialized = 1;
     }
 #endif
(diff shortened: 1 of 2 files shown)


larsbrinkhoff:
I'm wondering:

1. Is this change ok?  I have tested it with the KA10 simulator, and it seems to work just as well as using x11.c.

2. Is this the right place to submit this?  Or should it all go in Rich's repository?

larsbrinkhoff:
I have the corresponding KA10 changes in a pull request here: https://github.com/rcornwell/sims/pull/88

rcornwell:
Changes can go to my repository. However I would prefer all changes to SCP level code to also be mirrored to the main SimH tree. This way when I update the SCP layer I do not have to bother dealing with local additions. This also means that all changes are tested with other simulators which will cause less problems when KA10 and other simulators are merged into the main SCP tree.

PR #672, opened 2019-03-09
3b2: NI 10Base5 Ethernet Device

sethm:
Implements an Ethernet device ("NI", for "Network Interface") for the 3B2 simulator.


sethm:
`PDP11/pdp11_xq.c` was unfortunately a victim of automatic trailing white-space removal in my editor. I've removed that file from this commit, and it should be clean now.

PR #680, opened 2019-04-13
3b2: Soft-power shutdown via TIMER command

sethm:
This change enables the simulator to be shut down cleanly via a
soft-power shutdown command. This is implemented in the real 3B2/400
through the sanity timer, which, if it reaches zero, sets a bus timeout
flag in the CSR and issues an interrupt at IPL 15. The operating
system (System V UNIX) treats this as a shutdown request and enters
runlevel 0.

To use this change in a SIMH startup script, for example to implement a
3B2 simulator as a service, one could add these commands:

    # [... simulator setup ...]
    BOOT
    SET TIMER SHUTDOWN
    CONTINUE
    EXIT

On catching a SIGTERM, SIGINT, or SIGHUP, the simulator would return to
SCP control, set the soft power shutdown flag, and then continue
simulator execution. After the system is cleanly shut down, the
simulator would then exit back to the operating system.

### DIFF
--- a/3B2/3b2_cpu.c
+++ b/3B2/3b2_cpu.c
@@ -3455,7 +3455,7 @@ static void cpu_calc_ints()
         cpu_int_ipl = cpu_int_vec = CPU_ID_IF_IPL;
     } else if ((csr_data & CSRUART) || (csr_data & CSRDMA)) {
         cpu_int_ipl = cpu_int_vec = CPU_IU_DMA_IPL;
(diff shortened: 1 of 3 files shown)


markpizz:
This looks good.

Alternatively, the simulator startup could say:
```
# [... simulator setup ...]
SET ON
ON SIGINT SET TIMER SHUTDOWN; CONTINUE
BOOT
```

PR #685, opened 2019-04-24
3b2: Do not autoconfigure NI if attaching fails

sethm:
The NI device was autoconfiguring itself too early in the attach process, causing it to be in a bad state if attaching the device failed for any reason.

This change moves the NI device autoconfiguration to the very end of the attach routine, ensuring that the device is not visible to the system if attaching fails.


markpizz:
I have to admit to never having touched (or even seen) a 3b2 system, but from my experience with various other systems, whether a network adapter happens to be connected to a LAN shouldn't affect whether the device is configured or otherwise visible within an operating system.  On a real system someone might just happen to connect (or disconnect) the network adapter from the LAN at any arbitrary time.  ATTACHing a device is logically equivalent to connecting the network cable.

That said, I'll be glad to commit this change if you really want.  It is your simulator. :-)

PR #686, opened 2019-04-25
PDP11: fix SysID message data for LQA

pkoning2:
Correct the MOP data type code for the Device Type field in the SysID
message, and the value used for device type DELQA.


markpizz:
When do the values in the SysID messages come into play?

Are these part of a MOP load request or the related dialog?

I ask since I'd done some work to get these devices to succeed with a MOP boot at one time.  If you're changing anything that worked there, please test the changed code with various systems and describe what was done here.

Thanks.

pkoning2:
I wouldn't think so.  Load/dump is a separate protocol.  The SysID message is an informational message, intended to allow management stations to build a database of equipment on the network.
That said, I wouldn't mind doing some load tests, but I don't have any systems that do this.  Pointers would be welcome.

PR #703, opened 2019-05-16
Fixes for recent VAX simulators

9track:
- Core test is now passing for the VAX 8200. I've updated VAX_MINIMUM_DIAGS.DSK to include EBSAA which is required for the 8200 although the extended tests are failing so they are skipped for now.
- I found what appears to be a bug in the KA41-A ROM, which I had worked around previously by modifying the SIE in the cpu_boot function. This was incorrect and prevented the system from being able to boot.
- The video devices were not working on the VAXstation 3100 M76. This turned out to be an incorrect implementation of the configuration/test register.
- Coverity complaints from the recent run are fixed


9track:
The extended tests are also now passing for the VAX 8200. I've had to re-order vax-diag_test.ini a bit because EVKAD (compatibility mode instructions) cannot be run on this system.

PR #664, opened 2019-06-01
display: Type 342 lower case characters.

larsbrinkhoff:
I'm collecting information on the Type 342 lower case characters.  I will submit a pull request when I'm done.


larsbrinkhoff:
@markpizz this is now a pull request.

The characters have been updated to look like this, which I believe matches the evidence.
type342

The solid boxes are unknown characters.  The last line shows ~ followed by backspace and _, and use of sub and superscript.

markpizz:
Clearly you moved these commits into the sequence of commits from the master branch.  Can you do that over again and change the dates of to be more recent than 4 months ago...

larsbrinkhoff:
Those are the git commit "author dates" from when I started this, and those survive rebase operations.  I updated them to the current date as of today.

PR #720, opened 2019-07-10
KA10: Initial release of PDP10 KA/KI and PDP6 simulators.

rcornwell:
Initial release of PDP10 KA/KI and PDP6 simulators.


descrip.mms
```diff
@@ -38,10 +38,13 @@
...
-#            PDP8              Just Build The DEC PDP-8.
-#            PDP9              Just Build The DEC PDP-9.
-#            PDP10             Just Build The DEC PDP-10.
+#            PDP6            Just Build The DEC PDP-6.
+#            PDP7            Just Build The DEC PDP-7.
+#            PDP8            Just Build The DEC PDP-8.
+#            PDP9            Just Build The DEC PDP-9.
+#            PDP10           Just Build The DEC PDP-10.
+#            PDP10-KA        Just Build The DEC PDP-10 KA10.
+#            PDP10-KI        Just Build The DEC PDP-10 KI10.
```
larsbrinkhoff: Indentation changed?

PDP10 could now say "Just build The DEC PDP-10 KS10."

makefile
```diff
@@ -1869,6 +1878,60 @@ ifneq (,$(BESM6_BUILD))
     endif
 endif
 
+PDP6D = PDP10
```
larsbrinkhoff: = PDP6

descrip.mms
markpizz: It did change when all the new VAX model simulators.  I've fixed that now and added the KS10 wording.  Thanks.

makefile
markpizz: No.  The symbols ending with D are directory names.  All this new KA10, KI10 and PDP6 stuff are in the PDP10 directory with the original KS10.

markpizz:
Lars,

Rich forgot to pass along the latest update to display (type340.c).  I just included type340.c from his sims repo.  Meanwhile there is a remaining warning in that module:
```
display/type340.c:118:33: warning: '/*' within block comment [-Wcomment]
/*  ty340word DAC;              /* Display Address Counter */
```
The nested comment is easily addressed by putting */ before the nested /*, but this module has conditional code (on TYPE347) that actually references the DAC field.   Maybe the existence of this field should be conditional on TYPE347.  Please fix.

larsbrinkhoff:
It certainly looks that way.  All DAC references I see are inside TYPE347 conditionals.

I don't claim ownership of this code, but I'm willing to fix it.

CC @philbudne FYI

philbudne:
My suggestion/request is to put the DAC declaration under "#ifdef NOTYET" and remove the initial comment open.

The PC (DAC) doesn't depend on Type 347 (subroutine) support.  It would be required for Type 340C support (PDP-10 VB10C), *OR* I _think_ use of the 340 with any 18-bit CPU which I believe all allowed the 340 to request display instructions directly from memory.

Without prints, it's hard to know if the PDP-6 + Type 340 combination left out the DAC circuitry or somehow bypassed it, and made the 340 beg for instructions by interrupting the host.

larsbrinkhoff:
Fixed as per above in #721.

PR #721, opened 2019-07-10
Fix unbalanced comment in type340.c

larsbrinkhoff:
See here: https://github.com/simh/simh/pull/720#issuecomment-509915223


larsbrinkhoff:
Also fixed a line ending missing the CR.

PR #724, opened 2019-07-18
Travis matrix build.

larsbrinkhoff:
Split the monolithic Travis CI "make all" build into several jobs, one for each simulator.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,4 +15,4 @@ env:
   - SIM="uc15"
 sudo: required
 install: sh -ex .travis/deps.sh
-script: make $SIM
+script: make --jobs=4 $SIM
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,17 +2,6 @@ os:
   - linux
   - osx
 language: c
-env:
-  # These are supposed to match ALL in makefile.
-  # Each job builds 10 simulators.
-  - SIM="pdp1 pdp4 pdp6 pdp7 pdp8 pdp9 pdp10 pdp10-ka pdp10-ki pdp11"
-  - SIM="pdp15 vax microvax3900 microvax1 rtvax1000 microvax2 vax730 vax750 vax780 vax8200"
-  - SIM="vax8600 microvax2000 infoserver100 infoserver150vxt microvax3100 microvax3100e vaxstation3100m30 vaxstation3100m38 vaxstation3100m76 vaxstation4000m60"
-  - SIM="microvax3100m80 vaxstation4000vlc infoserver1000 nova eclipse hp2100 hp3000 i1401 i1620 s3"
-  - SIM="altair altairz80 gri i7094 ibm1130 id16 id32 sds lgp h316"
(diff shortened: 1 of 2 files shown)


markpizz:
Are these 'jobs' done sequentially, or in parallel?  If is done in parallel, then is there some way to have a separate job that is done first before the parallel jobs start.  Specifically, a job that does 'make BIN/buildtools/BuildROMs' is needed.  Otherwise, parallel jobs will race each other (and possibly clobber something) running this internal step.

larsbrinkhoff:
I believe they are done in parallel, subject to a limited number of virtual machines.

I also believe each job is done in a separate virtual machine, so they don't clobber each other.

larsbrinkhoff:
Well, all jobs finished.  Individual jobs are 1-7 minutes (16 for one outlier), but the accumulated time is over eight hours.  This is mostly because of the slow mac machines.

markpizz:
I feel guilty about the resources expended as a consequence of what might be a rather minor commit which often doesn't affect the code (and thus the build outcome).  For example: c1d9db8be7e47b31f

larsbrinkhoff:
That may be a concern.  Sometimes I wish there were a more selective approach to building commits, but it's all or nothing.

Maybe things will be different if a few simulators are bundled together in a job?  Let's try that and see what happens.

markpizz:
I'd say to go for 10 at a time.  Once this is stabilized, I'll change the make logic and the VAX test script to skip testing after more than 3 VAX simulators have run tests in a single make execution.

larsbrinkhoff:
I had each job build around seven simulators, and the build time was reduced to two hours.

I'll update it to do 10.

larsbrinkhoff:
With builds batched 10 to the job, the accumulated build time is around 1.5 hours.

markpizz:
Before making this the method, I'm wondering if/how well/how fast a "make -j" would execute and if it would correctly catch build errors.

markpizz:
Or maybe some limited -j like "-j 4" would be more reasonable...

markpizz:
or more clearly "--jobs=4"

markpizz:
I see you've got travis jobs with 10 simulators each and a make --jobs=4 going in each.  What I actually meant, was to see if, with one travis job, --jobs=4 runs quickly enough to avoid the extra travis overhead of setting up each new VM for the job and the configuration setup for each...

Also, for the sake of testing, try putting in some garbage at the beginning of one of the source modules (say PDP10) and see if the single travis job reports an error if the make has "--jobs=4" on it...

larsbrinkhoff:
Done.

Note: 98f1b80 failed because Travis stops a job if a new commit is made to the same branch.

markpizz:
From this message: "No output has been received in the last 10m0s, this potentially indicates a stalled "

It seems that --jobs=n isn't the right path to take.  I suspect that each VM is only given a single CPU so parallel compiles will just be thrashing.  Meanwhile the lack of continuous output during the thrashing (as indicated above) gets Travis confused anyway.

So, I think the separate jobs doing maybe 15 compiles each will be reliable.

markpizz:
Merged a while back by Cherry pick.

PR #746, opened 2019-09-16
Visual Studio CI build

larsbrinkhoff:
I'm experimenting with AppVeyor to add Windows continuous integration to SIMH.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 [![Coverity Scan Build Status](https://scan.coverity.com/projects/11982/badge.svg)](https://scan.coverity.com/projects/simh)
 [![Build Status](https://travis-ci.org/simh/simh.svg)](https://travis-ci.org/simh/simh)
+[![AppVeyor](https://ci.appveyor.com/api/projects/status/github/simh/simh)](https://ci.appveyor.com/project/simh/simh/history)
 
 ## Table of Contents:
 [WHAT'S NEW since simh v3.9](#whats-new-since-simh-v39)  
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,13 @@
+platform:
+  - x86
+cache:
+  - ../windows-build
+skip_commits:
+  files:
+    - doc/*
+    - '**/*.md'
+    - '**/*.txt'
+before_build:
+  "\"C:\\Program Files (x86)\\Microsoft Visual Studio 9.0\\VC\\vcvarsall.bat\""
+build_script:
+  cmd: cd "Visual Studio Projects" & vcbuild /M%NUMBER_OF_PROCESSORS% /useenv /rebuild Simh.sln "Release|Win32"


larsbrinkhoff:
Submitted pull request.

For this to work, you must login on AppVeyor and enable building this repository.  https://www.appveyor.com/  You can use the GitHub credentials for this.

markpizz:
Several comments (maybe less is more):
1) It's not 100% clear exactly how you're getting things into the test environment, but if you drive the build using vcbuild, AND if you've got git installed on the system (and in the current path), then you don't need to download the windows-build components as a zip file which then must be expanded.  Since you're not downloading a zip file, then you don't need 7zip to unzip the windows-build.zip file.  The VS2008 build logic will automatically fetch windows-build using git when it isn't there.
2) VS2008 only specifically can target x86 binaries and no x64.  These binaries (built with VS2008) run on all versions of Windows from XP onward.  x64 binaries have been tested and compared with x86 binaries running on x64 Windows and simulator performance difference is in the noise.
3) The MinGW build is specific to x86 (i.e. MinGW32 vs MinGW64).  Recalling the original history of why this was supported, just the minimal MinGW32 environment is desired and/or needed.  No install of Msys or extra Unix commands should exist in the path.  I just tested this now and starting with mingw-get-setup.exe, I click on "Install", then "Continue".  This ends up starting the GUI installation manager, where I ONLY select: mingw32-base-bin, and mingw32-gcc-g++-bin to be marked for installation and then click "Menu->Installation->Apply Changes" followed by Apply and eventually Close.  After this, the simh build_mingw.bat works for me.  Almost all simulators will compile and run when this has been done.

larsbrinkhoff:
Updated as per comments.

larsbrinkhoff:
Build log:

https://ci.appveyor.com/project/larsbrinkhoff/simh

markpizz:
This is certainly on the right track!

1) Since you're not specific about what "configuration" you are targeting in the vcbuild command line, it seems that it is building BOTH the Win32-Debug AND the Win32-Release versions of things.  This runs approximately twice as long as just running a single target with out much value gain.  The more rigorous compile is the Win32-Release case.  Doing only this would seem to be equivalent to the normal makefile build on other OSes.  Using the makefile, you have to explicitly specify DEBUG=1 on the make command line to generate debug targets.
2) Poking around the AppVeyor site it seems that you get 2 HyperThreads/Cores.  Given that, a parallel build would significantly speed up execution time.  There is logic in BuildAll.cmd to invoke vcbuild to do a parallel build based on the number of available CPUs.  If that logic can't be directly embedded in the AppVeyor configuration file, it certainly can be embedded in a Windows command file as is done in BuildAll.cmd.
3) I feel guilty about all of the work that flows out of pushing every commit to the github/simh/simh master branch and having that trigger the whole cascade of CI activities.  It seems that AppVeyor has support for a "Build cache" that can be leveraged to retain some things that don't change often (like the ../windows-build directory).  Similarly, if commits only affect doc/*, **/*.md, **/*.txt files, then builds it seems like the supported file filtering can be leveraged to not run a full CI build.

larsbrinkhoff:
With "Release|Win32" and /M2 the build time is down to 10 minutes.

I'm checking if %NUMBER_OF_PROCESSORS% can be used to determine parallelism.

larsbrinkhoff:
I think for appveyor.xml, we can simply use `vcbuild /M%NUMBER_OF_PROCESSORS%`.  It's set to 2.

larsbrinkhoff:
I added caching and skipping of commits.

markpizz:
> With "Release|Win32" and /M2 the build time is down to 10 minutes.

That's incredible.  The systems doing this must be much more powerful than those doing the Travis CI builds...

> I think for appveyor.xml, we can simply use vcbuild /M%NUMBER_OF_PROCESSORS%. It's set to 2.

If you look at the logic in BuildAll.cmd:
```
set _SIM_PARALLEL=8
if %_SIM_PARALLEL% GTR %NUMBER_OF_PROCESSORS% set _SIM_PARALLEL=%NUMBER_OF_PROCESSORS%
vcbuild /M%_SIM_PARALLEL% /useenv /rebuild Simh.sln "Release|Win32"
```
This logic puts an upper on bound parallelism of 8.

When developing BuildAll.cmd, parallelism initially started at 2, then it went to %NUMBER_OF_PROCESSORS% and ultimately ended up with this upper bound of 8.  On the systems I tested this on, more than 8 either slowed things down, caused things to come to a grinding halt or caused a less than nice failure.   vcbuild was developed when an environment with a NUMBER_OF_PROCESSORS never could exceed 8, so this limit was more than acceptable.

markpizz:
I don't think you can implement that logic directly in the appveyor.yml file, but you certainly could imbed that logic in a .cmd file that the appveyor.yml then invokes.

larsbrinkhoff:
I did see the logic, but I didn't consider it necessary to add it to the appveyor.yml file unless NUMBER_OF_PROCESSORS would suddenly increase which I would think is unlikely.  The yml file is pretty much guaranteed never to be used in any other context.

But if you insist, I can create a new cmd file for it.

markpizz:
I don't think we need it now, but someday AppVeyor may be more free with available processors in the free version or someone may pay for more and it would be nice if the lesson learned didn't have to be relearned.

Getting back to what it takes to actually run this automatically, exactly what should I do to make this actually work in the github/simh/simh master branch?

larsbrinkhoff:
You have to log in here: http://appveyor.com/  
You can use your GitHub credentials.

Then click Projects -> NEW PROJECT -> GitHub -> simh -> simh -> ADD.

PR #751, opened 2019-09-17
PDP11: Fix NG bugs.

### LINKED ISSUE #750: PDP-11 NG device has too many vectors
The NG device occupies 270-304 of the vector space:

```
sim> set cpu 11/45
Disabling XQ
sim> set ng enabled
sim> show ng
NG      type=unknown, scale=1, address=17764040-17764043
        vector=270-304, BR5
```

My intent was it should only have one vector: 270.  This is from pdp11_io_lib.c:

```
    { { "NG" },          1,  1,  0, 0,
        {04040}, {0270} },                              /* NG - vector display */
```

I thought setting numv, which is the second "1", would mean one vector slot.  What did I do wrong?

larsbrinkhoff:
Resolves #750 NG device has too many vectors.  This fix is needed for MIT's SITS operating system.

Also fixes wrong text.

### DIFF
--- a/PDP11/pdp11_ng.c
+++ b/PDP11/pdp11_ng.c
@@ -54,7 +54,7 @@ const char *ng_description (DEVICE *dptr);
 #define IOLN_NG   4
 DIB ng_dib = {
   IOBA_AUTO, IOLN_NG, &ng_rd, &ng_wr,
-  4, IVCL(NG), VEC_AUTO, {NULL}, IOLN_NG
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
I asked some time ago about SIMH supporting multiple display windows.  This is an actual use case.  The SITS binary I run is configured for five NG displays.  Hopefully I can assemble a new binary with just one.

PR #759, opened 2019-10-28
IBM1130: simplify code for string duplication.

ptroja:


markpizz:
This change is merely a 1 line change for the sake of change.  It doesn't solve a problem or add functionality.

If you come up with something more substantive, we'll be glad to embrace it.

Issue #665, opened 2019-02-12
PDP11 simulator in headless (batch) mode


shattered:
- #### Context

- #### the output of "sim> SHOW VERSION" while running the simulator which is having the issue

- #### how you built the simulator or that you're using prebuilt binaries

- #### the simulator configuration file (or commands) which were used when the problem occurred.

- #### the expected behavior and the actual behavior

- #### you may also need to provide specific pointers to data files that may be necessary to demonstrate the problem

I'm looking for a way to run automated tests with SIMH, specifically to compare output of RT-11 DIR command to output of RT-11 filesystem parser I've written, across a few dozen of disk images.

Looks like a command file with ATTACH, BOOT, SEND "DIR\r", BYE commands should work, but DIR command never reaches the simulated OS.   Is this expected to work?

markpizz:
What you want to do should be completely possible using the existing simh commands.

You will need more than the commands you list.

If you want specific help you'll have to provide the precise details of what you've tried and what happened when you did.

shattered:
It is, indeed, possible.  The trick is to SEND before BOOT:
```
set cpu 11/23 256k cis
set console wru=035
attach rl0 root-rt11v53.rl02
send -t after=5000000,delay=10000000,"DIR /VOL/FUL/BLO/COL:1 DL0:\r\035"
bo rl0
bye
```
Plus this patch, so that WRU char is also accepted from sent string:
```
diff --git a/scp.c b/scp.c
index 0f66f7b9..5396d181 100644
--- a/scp.c
+++ b/scp.c
@@ -12157,6 +12157,8 @@ if (snd && (snd->extoff < snd->insoff)) {               /* pending input charact
         char dstr[8] = "";
 
         *stat = snd->buffer[snd->extoff++] | SCPE_KFLAG;/* get one */
+        if ((*stat & 0177) == sim_int_char)
+            *stat = SCPE_STOP;
         snd->next_time = sim_gtime() + snd->delay;
         if (sim_isgraph(*stat & 0xFF) || ((*stat & 0xFF) == ' '))
             sprintf (dstr, " '%c'", *stat & 0xFF);
diff --git a/sim_console.c b/sim_console.c
index 2b7322af..73d8c6cc 100644
--- a/sim_console.c
+++ b/sim_console.c
@@ -2821,7 +2821,11 @@ t_stat sim_poll_kbd (void)
 t_stat c;
 
[...]

Issue #662, opened 2019-02-09
AltairZ80: Unable to boot from hdsk


jblang:
- #### Context
I'm trying to boot a CP/M 2.2 hard disk image downloaded from Peter Schorn's website.  Specifically i.dsk from the cpm2.zip archive.  I have run syscopy from the A: drive (cpm2.dsk) against the I: drive where the hard disk image was mounted. It ran successfully.

- #### the output of "sim> SHOW VERSION" while running the simulator which is having the issue
```
sim> show version
Altair 8800 (Z80) simulator V4.0-0 Current
    Simulator Framework Capabilities:
        32b data
        32b addresses
        no Ethernet
        Idle/Throttling support is available
        Virtual Hard Disk (VHD) support
        RAW disk and CD/DVD ROM support
        Asynchronous I/O support (Lock free asynchronous event queue)
        Asynchronous Clock support
        FrontPanel API Version 12
    Host Platform:
        Compiler: GCC 7.3.0
        Simulator Compiled as C arch: x64 (Release Build) on Feb  8 2019 at 11:19:54
        Memory Access: Little Endian
        Memory Pointer Size: 64 bits
        Large File (>2GB) support
        SDL Video support: No Video Support
        PCRE RegEx (Version 8.39 2016-06-14) support for EXPECT commands
        OS clock resolution: 1ms
        Time taken by msleep(1): 1ms
        OS: Linux cloud9 4.4.0-17763-Microsoft #253-Microsoft Mon Dec 31 17:49:00 PST 2018 x86_64 x86_64 x86_64 GNU/Linux
        git commit id: 66743dfb
        git commit time: 2019-02-07T11:45:48-0800
```
- #### how you built the simulator or that you're using prebuilt binaries
Built on Ubuntu from latest Github sources.

- #### the simulator configuration file (or commands) which were used when the problem occurred.

```
d tracks[0-7] 254
attach dsk0 cpm2.dsk
attach dsk1 app.dsk
attach hdsk0 i.dsk
;demo how to use other disk formats in CP/M 2
;it will mount as drive J:
attach hdsk1 appleiicpm.dsk
set hdsk1 format=apple-d2
set cpu 64k
set cpu noitrap
set cpu z80
set cpu altairrom
set cpu nonbanked
reset cpu
set sio ansi
set sio nosleep
boot hdsk
```

- #### the expected behavior and the actual behavior

I would expect the hard drive to boot CP/M, the same as cpm2.dsk.  Instead, I get the following:

```
[...]

Issue #661, opened 2019-01-25
Type 340 characters


larsbrinkhoff:
According to this 340 screen shot, the * character has a slightly different appearance from the one documented in DECUS NO. 7-13 340 DISPLAY PROGRAMMING MANUAL.

![340-asterisk](https://user-images.githubusercontent.com/775050/51752103-e8992c80-20b6-11e9-8f71-37fc036a9baa.png)

larsbrinkhoff:
And the comma looks different too.

![340-real](https://user-images.githubusercontent.com/775050/51752158-0ebecc80-20b7-11e9-8f00-99f4eee438a5.png)

larsbrinkhoff:
Simulator screen shot from https://github.com/simh/simh/pull/652#issuecomment-452197089

![340chars](https://user-images.githubusercontent.com/775050/50815060-0d9e4900-131c-11e9-8466-f7756ee59574.png)

larsbrinkhoff:
I'm closing this in favour of #664.

Issue #659, opened 2019-01-21
Error in scp.c in `_sim_debug_write_flush()`


sethm:
## Context

I've been using `sim_debug()` very heavily while working on a new feature, and I've encountered this crashing bug on several occasions.

The issue is in `scp.c` in `_sim_debug_write_flush()`, around line 12252.

```c
while ((eol = strchr (debug_line_buf, '\n')) || flush) {
    char *endprefix = strstr (debug_line_buf, ")> ");
    size_t linesize = (eol - debug_line_buf) + 1;
```

If the character `'\n'` does not appear in `debug_line_buf`, `strchr()` returns the null pointer, and causes the calculation for `linesize` to return a wildly incorrect value.

## `lldb` session

```
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0xffffffffffffffe1)
  * frame #0: 0x00007fff7421aed0 libsystem_platform.dylib`_platform_memmove$VARIANT$Haswell + 496
    frame #1: 0x0000000100052554 3b2`_sim_debug_write_flush(buf="", len=0, flush=1) at scp.c:12299
    frame #2: 0x0000000100044fdb 3b2`_sim_debug_flush at scp.c:12329
    frame #3: 0x0000000100044a81 3b2`run_cmd(flag=5, cptr="") at scp.c:7914
    frame #4: 0x00000001000012b2 3b2`sys_boot(flag=5, ptr="") at 3b2_cpu.c:625
    frame #5: 0x00000001000304ce 3b2`do_cmd_label(flag=0, fcptr="boot.ini", label=0x0000000000000000) at scp.c:3596
    frame #6: 0x00000001000265ef 3b2`do_cmd(flag=0, fcptr="boot.ini") at scp.c:3426
    frame #7: 0x00000001000248e0 3b2`main(argc=2, argv=0x00000001005001f0) at scp.c:2579
    frame #8: 0x00007fff74032ed9 libdyld.dylib`start + 1
    frame #9: 0x00007fff74032ed9 libdyld.dylib`start + 1
(lldb) f 1
frame #1: 0x0000000100052554 3b2`_sim_debug_write_flush(buf="", len=0, flush=1) at scp.c:12299
   12296	                    _debug_fwrite (countstr, strlen (countstr));
   12297	                    }
   12298	                debug_line_buf_last_endprefix_offset = endprefix - debug_line_buf;
[...]

PR #762, opened 2019-10-29
Run 'brew update' before installing packages.

larsbrinkhoff:
This is to work around a bug in Homebrew, but it may be a good idea anyway.

### DIFF
--- a/.travis/deps.sh
+++ b/.travis/deps.sh
@@ -1,6 +1,7 @@
 #!/bin/sh
 
 install_osx() {
+    brew update
     brew install sdl2
     brew install sdl2_ttf
     brew install vde


larsbrinkhoff:
The Travis build passes the deps.sh run on Mac, so I'd say this solves the problem.

markpizz:
Looking back over the last several Travis build logs for the simh/simh repo seems to show that sometimes there is a bunch of brew related output and the build fails, and sometimes it doesn't.

#659, #660 fail in the same way as #664, #665 and #667 while 2 in the middle #661 and #662 were successful.

larsbrinkhoff:
I should also post this link:  
https://travis-ci.community/t/syntax-error-unexpected-keyword-rescue-expecting-keyword-end-in-homebrew/5623

It points to a recent problem with Homebrew "auto update".  Workarounds suggested are to disable auto update, or do a manual update.  This pull request does the latter.

markpizz:
So, do you believe that this looks good now?

larsbrinkhoff:
Yes.

markpizz:
Merged with Cherry Pick, slight rephrased commit message, attribute Lars as author

PR #763, opened 2019-10-29
Add CI email notifications.

larsbrinkhoff:
Email messages should be sent to @markpizz when Travis CI builds fail, or are fixed, etc.

See this for details:  
https://docs.travis-ci.com/user/notifications

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,3 +14,8 @@ env:
 sudo: required
 install: sh -ex .travis/deps.sh
 script: make $SIM
+notifications:
+  email:
+    recipients:
+      - Mark@infocomm.com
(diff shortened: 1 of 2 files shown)


larsbrinkhoff:
I'm adding myself too.

The default behaviour is to always send a message when a build fails, and also when a failing build has been fixed.

larsbrinkhoff:
I'll let this hang around in draft state for a while so I can test failing builds.

larsbrinkhoff:
I think this should work.

markpizz:
Merged with Cherry Pick, slight rephrased commit message, attribute Lars as author

PR #778, opened 2019-12-11
TARBELL: Properly set current drive and drive not ready flag on reset

deltecent:
Fixes problem with current drive and drive not ready flags being set incorrect upon device reset. This was also causing "boot tarbell" to fail.


markpizz:
Merged with cherry pick.

PR #779, opened 2019-12-11
TARBELL: fix uptr NULL pointer before attach

deltecent:
Fixes possible NULL pointer when issuing commands to the controller before a disk has been attached.

### DIFF
--- a/AltairZ80/s100_tarbell.c
+++ b/AltairZ80/s100_tarbell.c
@@ -356,10 +356,6 @@ static t_stat tarbell_svc(UNIT *uptr)
     FD1771_REG *pFD1771;
     uint32 now;
 
-    if (uptr == NULL) {
-        return SCPE_IERR;
-    }
-
     pFD1771 = &tarbell_info->FD1771[tarbell_info->currentDrive];
 
     /*
@@ -383,10 +379,6 @@ t_stat tarbell_attach(UNIT *uptr, CONST char *cptr)
     char header[4];
     t_stat r;
     unsigned int i = 0;
-
-    if (uptr == NULL) {
-        return SCPE_IERR;
-    }
 
     r = attach_unit(uptr, cptr);    /* attach unit  */
     if(r != SCPE_OK) {              /* error?       */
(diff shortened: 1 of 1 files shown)


markpizz:
Unless you say you're really done.  I'm going to hold off on merging this until you do some more testing and collect any other changes.  I screwed up merging your other commits.  I intended to squash them all into a single commit that added the tarbell device, but was distracted by other things and broke my own rules.

markpizz:
There are couple of conventions we like to follow here:
1) Commits start with the affected simulator name (or global component) followed by a colon, with a summary of the commit up to 80 characters.  If extra commit message info is appropriate, a blank line follows the first line and then as much extra comments are needed with lines < 80 characters.
2) Prior to you issuing any commit, you should pull from the github.com/simh/simh master branch and then commit your change and generate a pull request.  If you do this, my merge process is simple and most of the time I can just pull your change(s) and then push it out to the master branch..  The commit rate on this project is very low (usually at most a couple of commits a day, often with days or weeks without any).  Therefore multiple commits from different folks are rare.

I'm specifically mentioning this now since your last pull request has already been merged, and yet this pull request still contains the commit from that request.  When I don't have to rewrite commit message and/or otherwise cherry pick a commit (and then remember to credit you as the author), things go much smoother.

markpizz:
I'm in no way suggesting that you shouldn't commit your work often, you can follow any pattern you want.  I'm just saying that if you commit often, you can probably collect any number of commits and then generate a pull request.  This lets you manage your own working environment however you want. 
 Then you can create a dedicated branch squash your own commit sets and pull from github.com/simh/simh master in preparation for a pull request - that pull request can be from a separate temporarily created branch dedicated to the particular pull activity.  After that request is merged you create a new one and go again.

AltairZ80/s100_tarbell.c
```diff
@@ -376,6 +384,10 @@ t_stat tarbell_attach(UNIT *uptr, CONST char *cptr)
     t_stat r;
     unsigned int i = 0;
```
markpizz: NULL pointer checks in the routines explicitly called by the simh framework (i.e. the DEVICE->attach=tarbell_attach, DEVICE->detach=tarbell_detach, UNIT->action=tarbell_svc) are meaningless.  In each of these contexts, uptr can't be NULL.

deltecent:
Going to hold off on this PR for now.

deltecent:
Mark, I completely understand and will do my best to follow these conventions.

Patrick

PR #783, opened 2019-12-16
AltairZ80: adds Tarbell SSSD disk interface

deltecent:
This version should be ready.


markpizz:
Patrick,

Several things:
1) in your code you seem to be assuming that the interval argument to sim_activate() is milliseconds.  What the interval value really is is instructions.  Many simulators schedule completion activities for devices using this paradigm and pick a minimal number of instructions that the original hardware would have been able to execute before the device completes the requested operation.  This minimal timing approach has devices which are very responsive when compared to the timing behavior of the original hardware, and since the processor for this simulator is running much faster than the original hardware ran everything is VERY snappy.  In order for this approach to make sense, your time reference needs to not be wall clock time, but the number of instructions executed.  Wall clock time is returned by sim_os_msec(), the number of instructions executed is returned by calls to sim_grtime() or sim_gtime().   If you really want the device to behave with timing characteristics that approximate the speed of the original hardware, then all of your timing activities should be using wall clock values, so sim_os_msec() with calls to sim_activate_after() is the approach you should take.  sim_activate_after() takes wall clock microseconds as its delay value.
2) I see you calling: sim_cancel() followed immediately by sim_activate().  The sim_activate_abs() achieves this to force scheduling based on instructions.  Similarly there is sim_activate_after_abs() to force scheduling based on wall clock time.
3) Your git pull request management isn't following the project standards I spelled out earlier:
> There are couple of conventions we like to follow here:
> 
> Commits start with the affected simulator name (or global component) followed by a colon, with a summary of the commit up to 80 characters. If extra commit message info is appropriate, a blank line follows the first line and then as much extra comments are needed with lines < 80 characters.

The file you're making changes to is s100_tarbell.c, which is a component of the AltairZ80 simulator, so the commit message should start with AltairZ80: with a summary of the changes being made and extra description as needed after a blank line.  Although TARBELL is important to you, it really isn't a global component.  :-)

> Prior to you issuing any commit, you should pull from the github.com/simh/simh master branch and then commit your change and generate a pull request. If you do this, my merge process is simple and most of the time I can just pull your change(s) and then push it out to the master branch.. The commit rate on this project is very low (usually at most a couple of commits a day, often with days or weeks without any). Therefore multiple commits from different folks are rare.
I also said:
>I'm in no way suggesting that you shouldn't commit your work often, you can follow any pattern you want. I'm just saying that if you commit often, you can probably collect any number of commits and then generate a pull request. This lets you manage your own working environment however you want.
Then you can create a dedicated branch, squash your own commit sets, and pull from github.com/simh/simh master in preparation for a pull request - that pull request can be from a separate temporarily created branch dedicated to the particular pull activity. After that request is merged you create a new branch and go again.  The temporary branch gets dropped and you start from the state in the master branch going forward.

Since, what you are asking me to merge are some minor changes you worked through to the same file, I'm looking for a single commit based on the current HEAD of the master branch containing these changes.  I can quickly merge pull requests submitted this way with just a couple of clicks.  If I've got to squash your commits and compose a summary message about what's been done I won't be in a hurry to merge it.

Let me know if you want to think through these points and make other changes and adjust your process before committing the current stuff.

deltecent:
Did I miss the documentation somewhere that explains all this?

markpizz:
The documentation about the API's can be found in the doc/simh.doc.  

This describes the specific API, but it doesn't cover the philosophical choice that a simulator developer gets to make regarding the goals of the simulator (or device simulator) that they're writing.  The choice is either 1) design something that will run all historical software that ran on or against the target simulator, or 2) design something that really feels like how the original system behaved while also running the original software.  "Feels" like here relates to wall clock timing.  The choice of which philosophy to take often depends on what kind of device is being implemented and how the original device "interacted" with a user.  Both of these choices can produce an accurate simulation.  Your choice of the different APIs you are using made it clear that you didn't understand how these things worked and thus mixed these different goals without understanding the consequences.  It was a lack of understanding/experience on your part, which I'm trying to educate you about.

I noticed these things since your pull request didn't follow the project's "simple to merge" pull request strategy and I had to dig into the scattered set of patches to come up with a single "squashed" change set and commit message.

After I first explained about the pull request "easy to merge" project philosophy, if you didn't understand precisely what I was talking about, you could have asked for details.  Since you didn't ask, but you once again presented a new pull request I had to look at what was happening with a microscope before I could merge it.

If you really want your latest change set merged into the master branch, I can do that, but given the mixed API set that I saw when I looked through things, it would seem that you might want to take a different stab at how you're doing things.

deltecent:
I thought I understood the merge process. I misunderstood using "TARBELL:" instead of "AltairZ80:"

I was mostly interested in the APIs. I've been trying to understand them by looking at the source and other software. I never did really understand the timer mechanisms. The only time I'm attempting to reproduce is 2 revolutions of the floppy disk before the head unloads. Not really important for anyone. But I wanted it because I'm writing a diagnostic tool (which is working great now, thanks to most of the work being done on the simulator).

I am certainly going to look at changing the timer work. I just therw numbers in there until it worked. I'm glad you pointed out the more proper way to do it.

I also work on ElvishArtisan/rivendell on GitHub. It took a while to learn their paradigms too.

Patrick

markpizz:
Useful info about using git to squash several commits:  https://stackoverflow.com/questions/6934752/combining-multiple-commits-before-pushing-in-git

Once your pull request has been merged into the master branch (either directly or by some local manipulation on my part), you pretty much always want to discard (delete) the branch you were working in for those changes.  Later on, you add new changes starting from the then current contents of the github.com/simh/simh/master branch.

deltecent:
Ah, I now see what you're getting at... yes, I will rebase and put in a correct commit comment.

Thank you for your patience.

Patrick

deltecent:
I now have the timer working as you specified. It works much better.

The AltairZ80 emulator has an option to sleep between keyboard input checks. This seems to impact the event timers. Do you know if there is a way for the Tarbell device to check if the SIO device Unit 0 has the UNIT_SIO_SLEEP option enabled and adjust the usec time accordingly?

Thank you,

Patrick

markpizz:
> The AltairZ80 emulator has an option to sleep between keyboard input checks. This seems to impact the event timers. Do you know if there is a way for the Tarbell device to check if the SIO device Unit 0 has the UNIT_SIO_SLEEP option enabled and adjust the usec time accordingly?

I don't actually.  The AltairZ80 simulator is Peter Schorn's (@psco) baby.  I'm pretty sure that this simulator existed before the simh framework had idle support (not waste host CPU cycles when the simulator knew that no useful progress would be made until something external to the simulator happened).  When idle support is leveraged by the simulator (calling sim_idle() during idle times), the simulator timing activities are very well behaved.  Any direct calling of sim_os_sleep() will interfere with the framework's management of timing activities.

markpizz:
You properly reorganized the commits against the master branch, so I closing this pull request.

The discussion about "idle" may be worth bringing up with @psco, but since what's there now has been there forever and it works, I'm not sure much reorganization is appropriate.

PR #789, opened 2019-12-27
B5500: Fixed DTC to fix error caused by SCP changes

rcornwell:


rcornwell:
I know why the change was put into place.

I can change the code to put out a ^S and ^Q if that is better.

markpizz:
My last comment was done from my phone, without a clear view of what was really there.

You avoid the NUL in the debug output by:

```
                 if (c1 != 0)
                     sim_debug(DEBUG_DATA, &dtc_dev,
                        "Datacomm transmit %d %02o %c\n", ln, c&077, c1);
```

How about sticking with logging every output character, since it might be useful:

```
                 sim_debug(DEBUG_DATA, &dtc_dev,
                         "Datacomm transmit %d %02o->0%o %c\n", ln, c&077, c1, isalnum(cl) ? cl : '.');

```

Why bother masking c with 077 in the debug statement, when you've already used the unmasked value in the line:
```
c1 = con_to_ascii[c];
```

rcornwell:
Try this change.

markpizz:
Squashed two commits with rebase.

PR #790, opened 2020-01-05
MS CRT deprecation: Use "ISO names" vice POSIX.

bscottm:
sim_iso_names.h provides the shim to rename functions that the MS CRT
considers "deprecated".  Added to future-proof the simh code base when
MS eventually decides to fully remove them from the CRT.

Can remove _CRT_NONSTDC_NO_WARNINGS from MSVC build.

(Close the pull request if not interested.)

### DIFF
--- a/AltairZ80/altairz80_sio.c
+++ b/AltairZ80/altairz80_sio.c
@@ -267,7 +267,7 @@ static void processDirEntry (const char *directory,
                              void *context) {
     if (filename != NULL) {
         NameNode_t *top = (NameNode_t *)malloc(sizeof(NameNode_t));
-        top -> name = strdup(filename);
+        top -> name = sim_strdup(filename);
         top -> next = nameListHead;
         nameListHead = top;
     }
--- a/Ibm1130/ibm1130_cpu.c
+++ b/Ibm1130/ibm1130_cpu.c
@@ -1436,7 +1436,7 @@ t_stat cpu_set_type (UNIT *uptr, int32 value, CONST char *cptr, void *desc)
     is_1800 = (value & UNIT_1800) != 0;                 /* set is_1800 mode flag */
 
     for (r = cpu_reg; r->name != NULL; r++) {           /* unhide or hide 1800-specific registers & state */
-        if (strnicmp(r->name, "XR", 2) == 0) {
+        if (sim_strnicmp(r->name, "XR", 2) == 0) {
             if (value & UNIT_1800)
                 CLRBIT(r->flags, REG_HIDDEN|REG_RO);
             else
--- a/Ibm1130/ibm1130_cr.c
+++ b/Ibm1130/ibm1130_cr.c
@@ -1321,7 +1321,7 @@ static t_bool nextdeck (void)
         if (*buf == '#' || *buf == '*' || *buf == ';')
             continue;                       /* comment */
 
-        if (strnicmp(buf, "!BREAK", 6) == 0) {  /* stop the simulation */
+        if (sim_strnicmp(buf, "!BREAK", 6) == 0) {  /* stop the simulation */
             break_simulation(STOP_DECK_BREAK);
             continue;
         }
@@ -1366,7 +1366,7 @@ static t_bool nextdeck (void)
 
             if (*tempfile == '\0') {                        /* first time, construct unique name */
                 strcpy(tempfile, "tempXXXXXX");             /* make a modifiable copy of the template */
-                if (mktemp(tempfile) == NULL) {
+                if (sim_mktemp(tempfile) == NULL) {
                     printf("Cannot create temporary card file name\n");
                     break_simulation(STOP_DECK_BREAK);
                     return 0;
@@ -1395,7 +1395,7 @@ static t_bool nextdeck (void)
                 fpos = ftell(deckfile);
                 if (fgets(buf, sizeof(buf), deckfile) == NULL)
                     break;                          /* oops, end of file */
-                if (buf[0] != '!' || strnicmp(buf, "!BREAK", 6) == 0)
+                if (buf[0] != '!' || sim_strnicmp(buf, "!BREAK", 6) == 0)
                     break;
                 alltrim(buf);
             }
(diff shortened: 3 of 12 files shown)


markpizz:
Let's hold off on "solving" the problem of MS removing one or more routines if/when it actually arises.  These C run time routines have been around for a very long time.  Some/all may have alternate implementations or slightly different "safer" APIs, but I believe that Microsoft has done all they ever will do (i.e. issue the warnings they do now and provide a reasonable way to suppress those warnings).  If you can cite an example of any vendor who has actually removed one of these (or similar) routines from the C specified run time library, we can talk about it.

You may have seen the apparent replacement of many/most/all of the ctype.h character classification routines as a possible example of why we should do this, but those routines were added when it became apparent that the MS provided routines were actually broken.  The implementation of the MS implemented <ctype.h> routines took the character argument and used it as an index into a table with attributes for each character.  The broken part was that the routines only worked when the character being examined was between 0 and 127.  Characters between 128 and 255 were working as negative indexes into this classification array and went off into the weeds.  We saw characters with greater than 127 values when some users came along and had non-ansi native character sets.  When the MS ctype.h problem was first noticed, an effort was initially made to try to change all the code that used these routines to be changed to explicitly call the locally implemented routines.  It became apparent that trying to change all the current and future code to call the sim_xxx implementations was going to be a tail chasing effort.  I realized that a better solution was to 1) include the native local ctype.h, 2) \#undefine any macro implementations of these routines and then 3) \#define a macro for each name to call the local implementation.  This works reliably since sim_defs.h \#includes <ctype.h> at the top to be sure that the C \#include guard for that routine will get defined and that will implicitly protect any macros which are defined locally in the event that a user \#includes <ctype.h> after \#including sim_defs.h.  When MS (or some other vendor) actually removes one of the routines you're addressing here, the same approach should be taken in sim_defs.h, but let's wait until it actually is needed.

bscottm:
`ctype.h` is an interesting corner case that has functionality issues. These, by contrast, are not functional changes, specifically, just renames to reduce deprecation warnings (which, obviously, can be turned off with the _CRT_NONSTDC_NO_WARNINGS manifest constant.) That notwithstanding, source should compile with the minimal set of warnings without manifest constants that silence them.

Rhialto:
On Mon 06 Jan 2020 at 08:06:48 -0800, Mark Pizzolato wrote:
> You may have seen the apparent replacement of many/most/all of the
> ctype.h character classification routines as a possible example of why
> we should do this, but those routines were added when it became
> apparent that the MS provided routines were actually broken.  The
> implementation of the MS implemented <ctype.h> routines took the
> character argument and used it as an index into a table with
> attributes for each character.  The broken part was that the routines
> only worked when the character being examined was between 0 and 127.
> Characters between 128 and 255 were working as negative indexes into
> this classification array and went off into the weeds.  We saw
> characters with greater than 127 values when some users came along and
> had non-ansi native character sets.  When the MS ctype.h problem was

Well *that* sort of implementation is actually very much allowed by the
relevant standards.  The ctype functions or macros, such as isalpha(),
are declared to take int parameters (to be compatible with history, I
presume). If you give them a char, it will be converted to int. But of
course the char may signed, hence "values between 128 and 255" don't
exist in that case. Instead you (most likely) get negative values and
any value differing from EOF lands you in the "undefined behaviour"
terrirory.

Here is a fragment from the NetBSD manual ctype(3):
```
STANDARDS
     These functions, with the exception of isblank(), conform to ANSI
     X3.159-1989 ("ANSI C89").  All described functions, including isblank(),
     also conform to IEEE Std 1003.1-2001 ("POSIX.1").

CAVEATS
     The first argument of these functions is of type int, but only a very
     restricted subset of values are actually valid.  The argument must either
     be the value of the macro EOF (which has a negative value), or must be a
     non-negative value within the range representable as unsigned char.
     Passing invalid values leads to undefined behavior.

     Values of type int that were returned by getc(3), fgetc(3), and similar
     functions or macros are already in the correct range, and may be safely
     passed to these ctype functions without any casts.

     Values of type char or signed char must first be cast to unsigned char,
     to ensure that the values are within the correct range.  Casting a
     negative-valued char or signed char directly to int will produce a
     negative-valued int, which will be outside the range of allowed values
     (unless it happens to be equal to EOF, but even that would not give the
     desired result).
```
n1570 (the C11 draft document, which is the freely available version
most close to the C11 standard; http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf )
says the same, more briefly:
```
7.4 [Character handling ]

1   The header <ctype.h> declares several functions useful for classifying and mapping
    characters.198) In all cases the argument is an int, the value of which shall be
    representable as an unsigned char or shall equal the value of the macro EOF. If the
    argument has any other value, the behavior is undefined.
```

Yes, it is annoying, but either one uses unsigned char everywhere, or
uses char with casts to unsigned char. Or one uses replacement
functions/macros with such a strategy.

-Olaf.
-- 
Olaf 'Rhialto' Seibert -- rhialto at falu dot nl
___  Anyone who is capable of getting themselves made President should on
\X/  no account be allowed to do the job.       --Douglas Adams, "THGTTG"

markpizz:
> Yes, it is annoying, but either one uses unsigned char everywhere, or
uses char with casts to unsigned char. Or one uses replacement
functions/macros with such a strategy.

All true, thus the replacement ctype functions are implemented and used via new macros pointing at these functions.  This is deliberate since these functions (macros) are used to act on user input which has proven to possibly included signed characters which have the high bit set.  I brought these into the discussion as an example of how/why replacement functions are sometimes justified in simh.

Meanwhile, this patch was about replacements for things like:
chdir(), unlink(), strdup(), mktemp(), mkdir(), fileno(), etc. which all work fine now but the MS compiler spits out a warning about being deprecated.  If any other system also warned about any of these, we might adopt a change, but these warnings are unique to MS and easily suppressed by compiling with _CRT_NONSTDC_NO_WARNINGS.  The choice is either compile with this \#define, or change many places in the code now and future submissions to avoid these warnings that can safely be ignored.  This to me isn't a change that fixes a problem, but one that possibly adds more problems migrating to the replacement APIs when future simulators and/or simulated devices are added.

PR #800, opened 2020-02-07
VIDEO: Add support for SDL2 game controllers

larsbrinkhoff:
As discussed in https://github.com/rcornwell/sims/pull/179

### DIFF
--- a/sim_video.c
+++ b/sim_video.c
@@ -37,11 +37,50 @@ t_bool vid_mouse_b1 = FALSE;
 t_bool vid_mouse_b2 = FALSE;
 t_bool vid_mouse_b3 = FALSE;
 static VID_QUIT_CALLBACK vid_quit_callback = NULL;
+static VID_GAMEPAD_CALLBACK motion_callback[10];
+static VID_GAMEPAD_CALLBACK button_callback[10];
+static int vid_gamepad_inited = 0;
+static int vid_gamepad_ok = 0; /* Or else just joysticks. */
 
 t_stat vid_register_quit_callback (VID_QUIT_CALLBACK callback)
 {
 vid_quit_callback = callback;
 return SCPE_OK;
+}
+
+static t_stat register_callback (void **array, int n, void *callback)
+{
+    int i, j = -1;
+
+    if (!vid_gamepad_inited) {
+        return SCPE_NOATT;
+        }
+
+    for (i = 0; i < n; i++) {
+        if (array[i] == callback)
+            return SCPE_ALATT;
+        if (array[i] == NULL)
+            j = i;
+        }
+
+    if (j != -1) {
+        array[j] = callback;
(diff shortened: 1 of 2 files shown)


larsbrinkhoff:
This adds to sim_video.h:

```
typedef void (*VID_GAMEPAD_CALLBACK)(int, int, int);
t_stat vid_register_gamepad_motion_callback (VID_GAMEPAD_CALLBACK);
t_stat vid_register_gamepad_button_callback (VID_GAMEPAD_CALLBACK);
```

This makes it possible to register functions to be called back when a joystick or game controller event happens.

A motion event will supply the unit, the axis, and the value.  Values are in the range -32678 to 32676.

A button event will supply the unit, the button, and the state.  The state is 1 for pushed, and 0 for released.

larsbrinkhoff:
Further development is planned: #799

markpizz:
I should have checked this earlier, but please fix this warning:

```
./sim_video.c: In function ‘void vid_controller_button(SDL_ControllerButtonEvent*)’:
./sim_video.c:1309:55: error: invalid conversion from ‘Uint8 {aka unsigned char}’ to ‘SDL_GameControllerButton’ [-fpermissive]
     b = SDL_GameControllerGetBindForButton (c, event->button);
                                                ~~~~~~~^~~~~~
In file included from /usr/include/SDL2/SDL_events.h:37:0,
                 from /usr/include/SDL2/SDL.h:41,
                 from ./sim_video.h:229,
                 from ./sim_video.c:30:
/usr/include/SDL2/SDL_gamecontroller.h:336:1: note:   initializing argument 2 of ‘SDL_GameControllerButtonBind SDL_GameControllerGetBindForButton(SDL_GameController*, SDL_GameControllerButton)’
 SDL_GameControllerGetBindForButton(SDL_GameController *gamecontroller,
 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

visible with:
```
    $ make GCC=g++ pdp11
```

markpizz:
Looking at the routine that the error I'm seeing is reported in and you've still got some compile time version checking going on.  The API's that the your game controller support are using were present in earlier versions of SDL2, but they just didn't work due to an internal error.  The checks against the runtime SDL2 version are more than sufficient, right?

larsbrinkhoff:
The one compile time check is there to avoid a compile time error, yet allow using SDL 2.0.0-2.0.3.

I believe the "game controller" API was added with SDL 2.0.0.  SDL1 only had the joystick API.

My code depends on SDL_GameControllerFromInstanceID being present, which was added in SDL 2.0.4.  If some were to compile against 2.0.3 or earlier, the compile time check will avoid a compile time error.  And the run time check will fall back to the joystick API.

larsbrinkhoff:
The fix for the g++ compile error is in #803.

PR #802, opened 2020-02-07
display: Add display_close.

larsbrinkhoff:
As per request.

Adds to display.h:

```
extern void display_close(void *dptr); 
```

Used to close a display and associated resources.  Should be called from SIMH device reset routines when DEV_DIS is off.


larsbrinkhoff:
Opened as a draft.  I will add calls to PDP11, PDP18b, PDP1, and TX-0 as appropriate.

PR #807, opened 2020-02-25
Display/video fixes.

larsbrinkhoff:
If there is more than one display device and they call display_close from their reset routines, they will conflict.  This adds a check that only the device that opened the display is allowed to close it.

The other commit is a cleanup as per SDL documentation.  Maybe it's harmless to initialize and quit both subsystems, but I'd rather err on the side of cautiousness here.

### DIFF
--- a/display/display.c
+++ b/display/display.c
@@ -359,6 +359,7 @@ static long queue_interval;
 #define Y(P) (((P) - points) / xpixels)
 
 static int initialized = 0;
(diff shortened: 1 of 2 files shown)


markpizz:
@larsbrinkhoff In the future, please include the rich explanation you provided above in the commit message as well as here.

Thanks.

larsbrinkhoff:
Ok.

PR #814, opened 2020-03-10
KA10: Add support for KL10A/B

rcornwell:
Add support for KL10A/B.


markpizz:
The First commit was merged directly.  The second was included as part of a number of pieces relating to build, doc, etc. with you as author.

PR #824, opened 2020-03-26
KA10: Corrected error in NIA Padding. Handle disabled RH devices.

rcornwell:
Regular update to KA10. Fixed incorrect handling of NIA20 Padding. RH devices now return correct status for disabled devices.


markpizz:
Merged with Cherry Pick

PR #826, opened 2020-03-26
3b2: Refactoring in preparation for Rev 3

sethm:
Refactor in preparation for the addition of a Rev 3 simulator for the
3B2/1000 system.

This change also includes a full cleanup of the rat's-nest of includes
and externs that plagued the 3B2 simulator and made it difficult to
understand and maintain. Headers are now required in the following
order:

  compilation unit -> "3b2_defs.h" -> {... dependencies ...}

Finally, HELP has been added to the CPU device.


markpizz:
These files don't have CRLF line endings:
   3b2_1000_defs.h
   3b2_400_defs.h
   rom_1000_bin.h

3b2.vcproj has references to:
   3b2_cpu.h
   3b2_mau.h
   3b2_mmu.h
   3b2_sys.h
   3b2_sysdev.h
and no references to any new include files.

In the makefile, I'm thinking that you're probably going to do the other 3b2 models in the same 3b2 directory, so it would seem that you should keep the ATT3B2D as the directory symbol for all of your files while you may indeed have model specific ATT3B2M400_OPT symbols.  Also, maybe like the vax which when the later models were added, we moved to model specific simulator names, that was also done for the vax (microvax3900) but since folks (or scripts) may exist for the original name, the build copies the microvax3900 binary to the original simulator name (vax).

sethm:
Hi Mark. I've pushed a new change to address the line endings, vcproj file, and makefile.

My plan is to rename the "3b2" target to "3b2-400" and introduce a new "3b2-1000" target alongside it when the time comes. I think you're right that the ATT3B2M400_OPT and ATT3B2M400D are not needed, I've moved those back to ATT3B2_OPT and ATT3B2D, respectively. They will be shared between both targets.

markpizz:
Sorry to nit pick, but I notice that the BuildROMs file that you're generating translates rom_1000.bin to rom_1000.bin.h.  The resulting file name has 2 '.'s.  All the others translate the original extension separator to an underscore.  Please fix and back out these commits and submit a single commit.  I will drop your repo from my local integration repo and re-add the remote to fetch your changed history.  If you change the generated name, then you should recapture the output of BuildROMs and remove line endings and commit that version.

sethm:
I believe this latest push has everything all in one commit, now, including the correctly named ROM file with correct line endings.

PR #835, opened 2020-04-14
AltairZ80: Client socket needs to be non blocking

psco:


markpizz:
This change is not necessary.  Look closely at the definition of sim_connect_sock in sim_sock.h.

Did you observe an actual problem or were you merely looking at other changes that had happened?

Please back out (change history) all of the recent commits in your psco/simh master branch to just before c1b4493 and then pull from simh/simh master to resynch the master branch in your repo.

Issue #838, opened 2020-04-27
ALTAIRZ80: Execution Logging (new feature)


deltecent:
- #### Context
The AltairZ80 simulator allows the logging of IN and OUT instructions using the `SET CPU DEBUG=LOG_IN` and `SET CPU DEBUG=LOG_OUT` commands.

I wanted a way to be able to log the execution of instructions between two addresses. What I have running now is `SET CPU DEBUG=LOG_EXEC` to enable the logging of each executed instruction and `SET CPU LOGEXEC=start/end` to set the start and ending addresses that will be logged (default=0000/FFFF).

The logged output is the same at the ``DDT``:

```
Debug output to "test.log"
Debug output to "test.log" at Mon Apr 27 13:32:34 2020
Altair 8800 (Z80) simulator V4.0-0 Current        git commit id: 580b3889
256 bytes [1 page] loaded at ff00.
CPU: C0Z0M0E0I0 A=7B B=0002 D=2CE9 H=FFFC S=0000 P=FF0C DCR C
CPU: C0Z0M0E0I0 A=7B B=0002 D=2CE9 H=FFFC S=0000 P=FF0D JNZ 0FF08h
CPU: C0Z0M0E0I0 A=BD B=0002 D=2CE9 H=FFFC S=0000 P=FF08 MOV A,M
CPU: C0Z0M0E0I0 A=BD B=0002 D=2CE9 H=FFFC S=0000 P=FF09 STAX D
CPU: C0Z0M0E0I0 A=BD B=0002 D=2CE9 H=FFFD S=0000 P=FF0A INX H
CPU: C0Z0M0E0I0 A=BD B=0002 D=2CEA H=FFFD S=0000 P=FF0B INX D
CPU: C0Z0M0E0I0 A=BD B=0001 D=2CEA H=FFFD S=0000 P=FF0C DCR C
CPU: C0Z0M0E0I0 A=BD B=0001 D=2CEA H=FFFD S=0000 P=FF0D JNZ 0FF08h
CPU: C0Z0M0E0I0 A=C9 B=0001 D=2CEA H=FFFD S=0000 P=FF08 MOV A,M
CPU: C0Z0M0E0I0 A=C9 B=0001 D=2CEA H=FFFD S=0000 P=FF09 STAX D
CPU: C0Z0M0E0I0 A=C9 B=0001 D=2CEA H=FFFE S=0000 P=FF0A INX H
[...]

PR #839, opened 2020-04-29
ALTAIRZ80: Added "SET/SHOW CPU HISTORY" and "HELP CPU"

deltecent:
This PR adds a CPU instruction history ring buffer that other emulators have.

It also adds "HELP CPU".

Reference #838


deltecent:
This should not have "HELP CPU" in it.

Issue #854, opened 2020-05-21
Issuing a step via remote telnet while simulator is running results in segfault


electrified:
Many thanks for simh, I am finding it very useful.

1. Altair Z80 simulator is running (not stopped on a breakpoint), with remote telnet configured
2. Via remote telnet issue a step command
3. Simulator segfaults

Expected result:
Step command should be ignored?

```
CP/M v3.0 [BANKED] on HBIOS v3.0.1

A>Remote Console Command from ::1> s

Program received signal SIGSEGV, Segmentation fault.
                                                    0x00007ffff7c07671 in fwrite () from /lib64/libc.so.6
(gdb) backtrace
#0  0x00007ffff7c07671 in fwrite () from /lib64/libc.so.6
#1  0x000055555566cfea in _debug_fwrite (buf=0x555556a310b0 "Samples are not being collected\n", len=32) at ./scp.c:12913
#2  0x000055555566d3c0 in _sim_debug_write_flush (buf=0x7ffffffe8240 "Samples are not being collected\n", len=32, flush=0) at ./scp.c:12975
#3  0x000055555566d6cf in _sim_debug_write (buf=0x7ffffffe8240 "Samples are not being collected\n", len=32) at ./scp.c:13020
#4  0x000055555566fc81 in Fprintf (f=0x0, fmt=0x5555556fcfc8 "Samples are not being collected\n") at ./scp.c:13614
#5  0x0000555555675d6f in sim_rem_sample_output (st=0x0, line=0) at ./sim_console.c:551
#6  0x000055555567a413 in sim_rem_con_data_svc (uptr=0x555556a25920) at ./sim_console.c:1732
#7  0x0000555555667230 in sim_process_event () at ./scp.c:11307
#8  0x00005555555a2841 in sim_instr_mmu () at ./AltairZ80/altairz80_cpu.c:2130
#9  0x00005555555a269c in sim_instr () at ./AltairZ80/altairz80_cpu.c:2033
#10 0x000055555565fecb in run_cmd (flag=4, cptr=0x7fffffff01e8 "") at ./scp.c:8527
#11 0x000055555564f2d7 in do_cmd_label (flag=0, fcptr=0x7fffffff9690 "Sim.cfg SBC_simh.rom", label=0x0) at ./scp.c:3906
#12 0x000055555564e3f0 in do_cmd (flag=0, fcptr=0x7fffffff9690 "Sim.cfg SBC_simh.rom") at ./scp.c:3730
[...]

PR #863, opened 2020-05-27
AltairZ80: Improved resource mapping diagnostics.

psco:


markpizz:
Is this related to the activities in #854 ?  If so, does it change anything that was observed there?

psco:
It does not change anything related to #854 but it can help to better understand what happens when a device is reset.

PR #882, opened 2020-06-14
IMLAC: Implement 8-level return stack for display processor.

larsbrinkhoff:
The Imlac display processor would only allow one single level of subroutine calls.  An option expanded this to an eight level stack.  This is needed by the Maze game.

### DIFF
--- a/imlac/imlac_dp.c
+++ b/imlac/imlac_dp.c
@@ -30,7 +30,8 @@
 #define DBG             0001
 
 static t_addr DPC;
-static t_addr DT;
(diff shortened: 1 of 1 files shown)


markpizz:
When you've got various activities that you work on on the side in separate branches and then eventually put them up as pull requests, please date them with current dates (or dates in sequence with the master branch if commits have been quiet) in the pull request commits.  This one is an example of that.

larsbrinkhoff:
I'll try to remember that.  Commit date updated.

PR #885, opened 2020-06-14
IMLAC: Fix bug in long vector instruction.

larsbrinkhoff:
X and Y were swapped.


larsbrinkhoff:
Maze now fully working.
maze-top
maze-inside

PR #890, opened 2020-06-16
APPVEYOR: Save built artifacts in a zip file.

larsbrinkhoff:
I'm not sure if this will work, but according to https://www.appveyor.com/docs/packaging-artifacts/, this should put the entire BIN directory in a zip file on the Appveyor web site.

The pull request is in a draft state until I can verify it works.  Also, is it a good thing to do?


larsbrinkhoff:
I do see a BIN.zip for download here: https://ci.appveyor.com/project/simh/simh/builds/33557392/artifacts

markpizz:
I see essentially no value for a accumulation of these (especially including all the build intermediate files in BIN/NT/Projects).  These aren't interesting to provide for every commit.  I already push out a set of binaries for Windows users who just want to get the binaries from time to time into the https://github.com/simh/Win32-Development-Binaries repo.

larsbrinkhoff:
I think the zip file could be restricted to the BIN/Win32-Release directory, but it seems you prefer your Win32-Development-Binaries solution so I'll close this.

PR #898, opened 2020-06-22
AltairZ80: adds "M2SIO0" and "M2SIO1" MITS 88-2SIO devices that support TMXR.

deltecent:
This PR adds two new devices, `M2SIO0` and `M2SIO1`. These two independent devices provide "raw" 88-2SIO ports that can be used instead of the `SIO` device for attaching socket and serial ports using TMXR.


psco:
Ok to merge, updated documentation with reference to sample software to follow.

PR #904, opened 2020-06-26
IMLAC: Three updates.

larsbrinkhoff:
- Implement CPU instruction execution history.
- Load "block" register when loading DPC.

Added:

- Remove unused variable.


markpizz:
Time is moving backwards again.  Please fix,   

This would be an opportunity to just fix the compiler warning we discussed the other day.

larsbrinkhoff:
Oh, right.  I'll fix that.

larsbrinkhoff:
Fixed.

PR #908, opened 2020-07-02
AltairZ80: Fixes console input race condition when both M2SIOx ports are enabled and unattached.

deltecent:


markpizz:
Submit this commit against the HEAD of the master branch and I'll accept it as long as the commit message is < 80 characters.

Extended commit message contents should wrap before 80 and be separated from the primary line by a blank line.

The tarbell commit, which can be part of the same pull request, but likewise should have a commit message that follows these rules.

If you make it part of the same commit, they will be in sequence and can be merged directly without separate merge commits which we avoid.

deltecent:
I am no longer submitting PRs. I'll just maintain my own fork.

markpizz:
Meanwhile, I was offline for more than a week due to both a physical move and a significant infrastructure build and thus missed your updates.

Fine with me keep away though. Good Luck.

PR #911, opened 2020-07-08
IMLAC: Fullscreen mode.

larsbrinkhoff:
This adds sim_video APIs for querying and setting fullscreen mode.  And then uses these APIs from the Imlac simulator.

When in fullscreen mode, the frame buffer is scaled to the size of the display and positioned in the center.

### DIFF
--- a/sim_video.c
+++ b/sim_video.c
@@ -1404,6 +1404,20 @@ if (SDL_SemWait (vid_mouse_events.sem) == 0) {
     }
 }
 
+t_bool vid_is_fullscreen (void)
+{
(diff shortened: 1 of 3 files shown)


sim_video.c
```diff
@@ -1418,21 +1418,38 @@ else
...
+}
+
 void vid_update (void)
 {
 SDL_Rect vid_dst;
-
-vid_dst.x = 0;
-vid_dst.y = 0;
-vid_dst.w = vid_width;
-vid_dst.h = vid_height;
```
larsbrinkhoff, line 1414: `vid_dst` was previously unused in the code below.

PR #912, opened 2020-07-12
AltairZ80: Change how PMMI device monitors socket connections.

deltecent:
Change PMMI to rely on CTS modem input status to determine if there is a socket connection rather than tmxr_poll_conn(). Also change default RTS setting to follow DTR.

### DIFF
--- a/AltairZ80/s100_pmmi.c
+++ b/AltairZ80/s100_pmmi.c
@@ -183,9 +183,9 @@ static MTAB pmmi_mod[] = {
     { MTAB_XTD|MTAB_VDV,    0,                      "IOBASE",  "IOBASE",
         &set_iobase, &show_iobase, NULL, "Sets MITS 2SIO base I/O address"   },
     { UNIT_PMMI_RTS,       UNIT_PMMI_RTS,     "RTS",    "RTS",    NULL, NULL, NULL,
-        "RTS follows DTR" },
+        "RTS follows DTR (default)" },
     { UNIT_PMMI_RTS,       0,                  "NORTS",  "NORTS",  NULL, NULL, NULL,
-        "RTS does not follow DTR (default)" },
+        "RTS does not follow DTR" },
     { MTAB_XTD|MTAB_VDV|MTAB_VALR,  0,   "BAUD",  "BAUD",  &pmmi_set_baud, &pmmi_show_baud,
         NULL, "Set baud rate (default=300)" },
     { 0 }
@@ -194,7 +194,7 @@ static MTAB pmmi_mod[] = {
 static PMMI_CTX pmmi_ctx = {0, 0, PMMI_IOBASE, PMMI_IOSIZE, 0, pmmi_tmln, &pmmi_tmxr, PMMI_BAUD, 1};
 
 static UNIT pmmi_unit[] = {
-        { UDATA (&pmmi_svc, UNIT_ATTABLE | UNIT_DISABLE, 0), PMMI_WAIT },
+        { UDATA (&pmmi_svc, UNIT_ATTABLE | UNIT_DISABLE | UNIT_PMMI_RTS, 0), PMMI_WAIT },
 };
 
 static REG pmmi_reg[] = {
@@ -358,6 +358,11 @@ static t_stat pmmi_svc(UNIT *uptr)
 
         /* Enable receiver if CTS is active low */
         xptr->tmln->rcve = !(xptr->ireg2 & PMMI_CTS);
+
+        /* If socket, connection status follows CTS */
+        if (!xptr->tmln->serport) {
+            xptr->conn = !(xptr->ireg2 & PMMI_CTS);
+        }
     }
 
     /* TX data */
@@ -746,11 +751,11 @@ static int32 pmmi_reg3(int32 io, int32 data)
             if (xptr->oreg0 & PMMI_SH) {
                 xptr->ireg2 &= ~PMMI_AP;   /* Answer Phone Bit (active low) */
             }
-            sim_debug(STATUS_MSG, &pmmi_dev, "set DTR HIGH s=%04X.\n", s);
+            sim_debug(STATUS_MSG, &pmmi_dev, "set DTR HIGH.\n");
         } else {
             tmxr_set_get_modem_bits(xptr->tmln, 0, s, NULL);
             xptr->ireg2 |= PMMI_AP;
-            sim_debug(STATUS_MSG, &pmmi_dev, "set DTR LOW. s=%04X\n", s);
+            sim_debug(STATUS_MSG, &pmmi_dev, "set DTR LOW.\n");
         }
     }
 


markpizz:
> The problem is what happens when PMMI raises
DTR. In this scenario, tmxr_poll_conn() will no longer accept the
connection, tmxr_set_get_modem_bits() will and there is no way for
the DEVICE to know that a connection was actually accepted at
the time DTR is set or in the _svc routine. The only way to know
that there is now an active connection is to look directly at
tmln.conn.

Not true.  Once you've raised DTR, you can then call tmxr_set_get_modem_bits() again, and at that point if the line is connected, DSR will be set and DCD will also be set if you have set CTS.  Neither DSR or DCD will be set if there is no connection.

deltecent:
I didn't think it was possible to set CTS as that's in input which is set by TMXR. Did you mean RTS?

So what you're saying is:
```
tmxr_set_get_modem_bits(raise DTR);
tmxr_set_get_modem_bits(get incoming status);
connected = DSR | DCD status;
```
That seems like a better solution if that's what you're getting at.

markpizz:
Actually, I misspoke.  if you've set RTS (as well as DTR), then CTS will be set AND then DCD will also be set.

deltecent:
I'm glad you brought this up. I was fixated on using tmln.conn for knowing the connection status, not modem status lines. And since the PMMI only has DTR, RI, and CTS, I wasn't paying much attention to the others. I'll give this other way a try and report back. I also have the option to have RTS follow DTR, and the default is off. I may change the default to have RTS follow DTR.

markpizz:
Your logic treats a socket connection differently from a serial port.  Why is that?  The point is for devices to actually be unaware of the physical connection issues.

deltecent:
> Your logic treats a socket connection differently from a serial port. Why is that? The point is for devices to actually be unaware of the physical connection issues.

A serial port will work just fine without DTR or RTS. Some S100 devices only have DTR, some only have RTS, some have RI and CTS, some have DCD and DSR, some use CTS for DCD, some don't have RI or DCD. With TMXR just passing through these signals with a serial port, the device can use them, manipulate them, or ignore them. I don't see that flexibility with TMXR and sockets. TMXR assumes that DTR must be present to accept a socket connection, for example. Now if there's a way to treat sockets with passthru like a serial port (just tell me when there's a connection and when it goes away), then all the modem signals could be emulated by the device rather than TMXR, but I have not discovered a way to do that.

I have not dug into this too deeply, but in another case, it looks like TMXR raises DTR and RTS when the serial port is attached. The device wants them to default to off because they shouldn't turn on until the software tells them to, so when the serial connection comes in from tmxr_poll_conn, they're dropped, and I can still transmit and receive data. If I do that with a socket, it won't accept the socket connections, so data can be sent/received, so I only drop DTR/RTS if it's a serial port. TMXR is emulating serial port signals a specific way. Not all simulated devices work that way.

As before, I am not a TMXR expert. I didn't write it and the documentation does not state how much of it works. Much of it is trial and error and the source code is written in such a way that is hard for me to read (I like my opening and closing braces lined up, for example), so trial and error is sometimes easier than figuring out the code.

If you know a better way of doing this, I'm always eager to learn from the authors.

markpizz:
I'm not sure i follow your thinking. 

When connecting to physical serial ports, your simulator sets the modem bits you can set and it gets the incoming bits from the physical serial port.

Things logically plugged into serial ports are DCE devices (Modems), or if they're not they connect via a "NULL Modem" wiring situation which presents enough stuff for the port to work.  Often such connections present DSR (and/or DCD) on.  When these signals are present on the serial port connector, then TMXR will make them visible to the simulated device.  In the case of socket connections, the simulated modem signals follow the behaviors of how DCE would behave.

If it is important to present connection state into the simulated system by the device, then setting DTR (possibly all the time) and observing the DSR DCD signals passes that info inward.  If, on the other hand, the simulated system really doesn't care about actual connection, then you can send output to the port all day long and just like with a physical serial port, that output will collect on the floor in the back of the computer.  Since these output bits have no dimension, they collection behind the computer could go on infinitely.  :-)

deltecent:
> In the case of socket connections, the simulated modem signals follow the behaviors of how DCE would behave.

A DCE In 1976 or 2020? That's the issue. There was no standard behavior in 1976 (the MITS 88-2SIO works different than the PMMI modem). TMXR is attempting to emulate modem signals for a communications device (socket) that has no modem signals, thus they have to be emulated based on assumptions that may not work when emulating various serial devices, especially when emulating devices built in the 1970s. That is the reason TMXR itself must treat a socket connection different from a serial port, because they are different. I can jumper DCD on my modem to DSR on the host, or to CTS, or to CTS and DSR, or maybe just RI, raise DTR or not raise DTR, depending on what the emulated hardware and software need, and TMXR doesn't care because the tmxr_set_modem_control_passthru()` seems to cause TMXR to "get out of the way" when using a serial port, but not when using a socket. That is why the device must be aware of the physical connection because TMXR behaves differently when using a socket than when using a serial port. If one of TMXR's socket assumptions gets in the way, the device needs to do what it can to get around that assumption. An emulated serial device that's attached to a serial port doesn't need to worry about TMXR's assumptions because there aren't any. The input and output signals are physical, not emulated, and TMXR doesn't seem to care if DTR or high or low, it will still send/receive data or do whatever it is the device tells it. That's not the case when using a socket. If TMXR actually behaved the same when using a socket or serial port, then yes, the device wouldn't need to care, but that's not how TMXR appears to work.

markpizz:
> > In the case of socket connections, the simulated modem signals follow the behaviors of how DCE would behave.
>
> A DCE In 1976 or 2020? That's the issue. There was no standard behavior in 1976 (the MITS 88-2SIO works different than the PMMI modem). 
> 
> TMXR is attempting to emulate modem signals for a communications device (socket) that has no modem signals, thus they have to be emulated based on assumptions that may not work when emulating various serial devices, especially when emulating devices built in the 1970s. 

Actually, it was specifically designed to behave line DTE/DCE interactions from the 1970's, and it works well with many devices.

> That is the reason TMXR itself must treat a socket connection different from a serial port, because they are different. I can jumper DCD on my modem to DSR on the host, or to CTS, or to CTS and DSR, or maybe just RI, raise DTR or not raise DTR, depending on what the emulated hardware and software need, and TMXR doesn't care because the tmxr_set_modem_control_passthru()` seems to cause TMXR to "get out of the way" when using a serial port, but not when using a socket. 

It is not clear what you would want when using a socket.  You can transmit all day on a socket based line whether it is connected or not just like you can do with a physical serial port.  You can read from a disconnected socket all day and just like a serial port without anything plugged into it, you won't get anything.

When and why does your DEVICE need to know if a line is connected?

> That is why the device must be aware of the physical connection because TMXR behaves differently when using a socket than when using a serial port. If one of TMXR's socket assumptions gets in the way, the device needs to do what it can to get around that assumption. An emulated serial device that's attached to a serial port doesn't need to worry about TMXR's assumptions because there aren't any. The input and output signals are physical, not emulated, and TMXR doesn't seem to care if DTR or high or low, it will still send/receive data or do whatever it is the device tells it. That's not the case when using a socket. If TMXR actually behaved the same when using a socket or serial port, then yes, the device wouldn't need to care, but that's not how TMXR appears to work.

> One of the assumptions I'm trying to get around is TMXR not accepting an incoming socket connection, rather only raising RI and queuing the connection, if DTR is low. The assumption is DTR is to be used for flow control whether the device wants it to or not and an incoming connection will be used to drive RI whether the DEVICE wants it to or not, all of which could be done in the DEVICE if the author chose. This behavior does not exist when using a serial port.

Well, TMXR, for socket connections, is emulating a modem.  Modems in 1976 DID NOT answer incoming calls unless the connected device presented DTR.  If software running in the simulated system wants to support arbitrary connections from modems, then it can interact with the RI sequencing, or leave DTR turned on all the time.  Supporting this should be up to the software running inside the simulator.  If that software supported modem connections then it will behave this way.  Please point to a specific reference that describes a modem that was sold that used DTR for flow control.  Sure, some random piece of hardware may have done this, but that's not a candidate to connect via TMXR's socket interface.  If you want to connect such a device, us a serial port.

> > ...then you can send output to the port all day long and just like with a physical serial port, that output will collect on the floor in the back of the computer

> I don't believe that statement to be true. All a serial port needs is TX, RX, and Ground, unless the DEVICE and/or SOFTWARE say otherwise. A serial port isn't a logical connection like a socket. A serial port's connection is always physical. Bits won't fall on the ground like they will if TMXR doesn't handle an incoming socket connection. 

If nothing is connected to a physical serial port, were do transmitted bits go when they are sent out that port?  I believe the fall on the ground.  Transmitted bits through TMXR behave the same way if nothing is connected to either a configured serial port or a socket defined link.

> The RS-232 signals on the wire don't care and TMXR doesn't impose its will on a DEVICE when using a serial port, only when using a socket, essentially making the DEVICE and all the 40+ year old software be redesigned, rewritten or patched to conform to TMXR, but only when using a socket.

What 40+ year old software talked to modems, and didn't raise DTR?

markpizz:
Looking more closely at the description you've got at the top of the s100_pmmi.c module, you say:

>PMMI Communications MM-103 Modem & Communications
   adapter.

Is this a modem, or is it a serial port you could connect random devices to?  It would seem that it must be one or the other, but not both.

If it is a modem, then the DEVICE needs to interact with TMXR as TMXR needs and not necessarily expose those details to the simulated system.

deltecent:
I am no longer submitting PRs. I'll just maintain my own fork.

PR #913, opened 2020-07-13
AltairZ80: Fix SS1 device to return correct 2-digit year.

deltecent:
The SS1 device was returning (tm.tm_year-22) for the 2 digit year
causing the current year to be returned as '98' (120-22=98). This
PR changes the year calculation to (tm.tm_year%100).


markpizz:
@psco absolutely needs to agree with this one.  This change returns a year which most programs that ran on this system would likely be surprised by.

PR #929, opened 2020-09-26
Fixed issue with Nova carry flag not showing CPU history

Quantx:
Problem: Currently when viewing the Nova CPU history the carry flag is always displayed as 0 regardless of it's actual value at the time.

Cause: The carry bit is stored in `bit 17` and is lost when stored into the `int16 carry` member of `struct Hist_entry`

Solution: Shift carry into `bit 0` before storing it in the CPU history. Improve code for displaying carry bit.

### DIFF
--- a/NOVA/nova_cpu.c
+++ b/NOVA/nova_cpu.c
@@ -1316,7 +1316,7 @@ if ( hist )
     hist_ptr->ac1   = AC[ 1 ] ;
     hist_ptr->ac2   = AC[ 2 ] ;
     hist_ptr->ac3   = AC[ 3 ] ;
-    hist_ptr->carry = C ;
+    hist_ptr->carry = C >> 16 ;
     hist_ptr->fp    = FP ;
     hist_ptr->sp    = SP ;
     hist_ptr->devBusy    = dev_busy ;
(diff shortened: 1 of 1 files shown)


markpizz:
This is a reasonable change.

1) The simh project conventions, due to the low commit rate, are generally to avoid merge commits in the repo master branch.  This is generally achieved by doing a pull from the master branch BEFORE your commit (and push to your repo) followed immediately by generating a pull request and/or an email notifying me of your actions.
2) The simh commit message convention is to format commit messages as:  
SIMULATOR/COMPONENT: \<80 char summary text\>
\<blank-line-if-additional explanation is appropriate\>
\<optional-additional explanation wrapped at less than 80\>

Therefore an appropriate commit message for this change might be:
NOVA: Fix carry display in CPU instruction history

Additionally, Some part of your local git setup has your email text name as Quantx rather than "Samuel Deutsch" which then causes Quantx to be the committer name visible in the resulting commit.  Keep this if you want, but most folks usually appreciate name recognition.

I can cherry pick this commit and commit it with you as author, with the suggested commit message, or you can drop this pull request and rework it with the above feedback.

By the way, no harm in including the whole problem description and solution you describe in this pull request as optional additional explanation in the actual commit message.  Then the details are represented in all instances of the repository.

Quantx:
I think I'll close this pull request and open a new one with the changes you've recommended. Thanks for your help.

PR #932, opened 2020-10-06
Enable warnings in build

promovicz:
I build simh with custom compiler options. One of the things I'm doing out of habit is enabling -Wall.

It used to be that simh wasn't to bad about it. A warning here and there. But this has changed dramatically since I last checked.

Building simh with just -Wall (no -Wextra, which is less portable in its effect) creates a thunderstorm of warnings, many of which look like they are surefire bugs.

It is my belief that - in the interest of everybody - warnings should be enabled by default and also in travis builds so that this can be fixed incrementally.


markpizz:
Rather than changing the makefile, if your efforts compiling with different compilers and/or compiler options actually illuminate true bugs, or raise questions about some things (which you mention in passing as surefire bugs), PLEASE create issues which suggest that there is a problem and, if possible, offer suggestions about how to fix each of them.

The change you suggest here is effectively equivalent to:
```
$ make GCC="gcc --Wall"
```
and thus should suffice for the situations you may try to create.

promovicz:
I disagree. These warnings need to be visible to everyone so that they can be fixed. Otherwise they will just accumulate and continue to produce subtle bugs.

Can this please be given some serious consideration?

I am perfectly willing to reformulate my patch however you want, but I strongly believe in the bug fixing power of warnings and would really appreciate if we could work together to do something about this.

promovicz:
Also: I am perfectly willing to hunt down the actual warnings that pop up when this gets enabled. I just think that this needs to be a community effort, and enabling warnings is the first step to achieving that.

It works well for most other packages in the world. The standard seems to be "-Wall everywhere" and "-Wextra for developers only". Maybe simh should follow?

markpizz:
After the plethora of noise produced by -Wall (most of  which is actually not operational) comes down to a murmur in the simh codebase, we can consider such a change to the base makefile.

I didn't in anyway intend to suggest that problems need not be fixed, but every warning is not a problem, especially for this codebase which compiles the same modules for different simulators with different build options and thus the same module may emit warnings for some but not others.

promovicz:
I am particularly worried about the enormous prevalence of unitialized variable use.

But fair enough. Fix warnings first, enable them later.

As for "not every warning is a problem" I don't agree. Code isn't maintained by computers but by humans. When I patch simh I want to submit only code that is warning-free, and that is impossible with the current mess.

markpizz:
> I am particularly worried about the enormous prevalence of unitialized variable use.

These warnings are a feature of recent gcc functionality.  They may actually be true, or not.  The compiler doesn't seem to be doing as deep a code analysis as the Coverity scans do.  We've aggressively pursued issues that those scans point out about issues like this.

rcornwell:
Doing a build with -Wall results in a whole bunch of stupid warnings in sim_debug macro when &dev is used instead of dptr!

promovicz:
> > I am particularly worried about the enormous prevalence of unitialized variable use.
> 
> These warnings are a feature of recent gcc functionality. They may actually be true, or not. The compiler doesn't seem to be doing as deep a code analysis as the Coverity scans do. We've aggressively pursued issues that those scans point out about issues like this.

Cool. That actually alleviates my worries somewhat, although maintainability still suffers.

I suppose this is for another day then. Feel free to close unless you want to keep this discussion for reference.

markpizz:
The noise level of warnings has been reduced to a minor level when building with -Wall, but we're not going to go with -Wall as a default since there are a good number of cases where variables and/or routines are defined but sometimes not used depending on particular simulator feature needs.

This discussion will remain for reference even as a closed pull request.

PR #933, opened 2020-10-07
AltairZ80: Adds support for the Tarbell 2022 double density controller

deltecent:
The 'tarbell' device now supports Tarbell model 1011 single density
and model 2022 double density floppy disk controllers.

The model is selected using the "SET TARBELL MODEL={SD|DD}" command.


psco:
This looks good and should be included.

markpizz:
Hi Patrick,

Please reformat your commit message to follow this pattern:

```
<affected-component>: 1 line summary description of change < 80 in total length
<blank-line-if-extra-commit-message-is-appropriate>
<Extra commit message content wrapped with lines less than 80>
```

You might as well remove the unused variables while you're at it.

I could do both and commit with you as author, but some folks prefer to get credit for commits...

PR #956, opened 2020-11-10
sim_imd: Add missing format specifier in debug print.

hharte:


markpizz:
I solved this a while back by explicitly casting the ftell() result to an (int) which certainly should be fine for any floppy disk image which sim_imd is targeted at.

PR #935, opened 2020-11-14
VIDEO: Support for multiple displays

larsbrinkhoff:
I'm working on a simulation of the General Turtle TT2500.  As can be seen below, it had two separate displays: one for text, and another for vector strokes.  I wonder if extending the SIMH video API to multiple displays would be under consideration?

TT2500

### DIFF
--- a/scp.c
+++ b/scp.c
@@ -1567,9 +1567,7 @@ static const char simh_help2[] =
       "+sh{ow} ethernet             show ethernet devices\n"
       "+sh{ow} serial               show serial devices\n"
       "+sh{ow} multiplexer {dev}    show open multiplexer device info\n"
-#if defined(USE_SIM_VIDEO)
       "+sh{ow} video                show video capabilities\n"
-#endif
       "+sh{ow} clocks               show calibrated timer information\n"
       "+sh{ow} throttle             show throttle info\n"
       "+sh{ow} on                   show on condition actions\n"
@@ -2416,7 +2414,7 @@ static const char simh_help2[] =
        /***************** 80 character line width template *************************/
 #define HLP_SCREENSHOT  "*Commands Screenshot_Video_Window"
       "2Screenshot Video Window\n"
-      " Simulators with Video devices display the simulated video in a window\n"
+      " Simulators with Video devices display the simulated video in a window(s)\n"
       " on the local system.  The contents of that display can be saved in a\n"
       " file with the SCREENSHOT command:\n\n"
       "++SCREENSHOT screenshotfile\n\n"
@@ -2546,9 +2544,7 @@ static CTAB cmd_table[] = {
     { "SLEEP",      &sleep_cmd,     0,          HLP_SLEEP,      NULL, NULL },
     { "!",          &spawn_cmd,     0,          HLP_SPAWN,      NULL, NULL },
     { "HELP",       &help_cmd,      0,          HLP_HELP,       NULL, NULL },
-#if defined(USE_SIM_VIDEO)
     { "SCREENSHOT", &screenshot_cmd,0,          HLP_SCREENSHOT, NULL, NULL },
-#endif
     { "TAR",        &tar_cmd,       0,          HLP_TAR,        NULL, NULL },
     { "CURL",       &curl_cmd,      0,          HLP_CURL,       NULL, NULL },
     { "RUNLIMIT",   &runlimit_cmd,  1,          HLP_RUNLIMIT,   NULL, NULL },
@@ -2640,9 +2636,7 @@ static SHTAB show_glob_tab[] = {
     { "SERIAL",         &sim_show_serial,           0, HLP_SHOW_SERIAL },
     { "MULTIPLEXER",    &tmxr_show_open_devices,    0, HLP_SHOW_MULTIPLEXER },
     { "MUX",            &tmxr_show_open_devices,    0, HLP_SHOW_MULTIPLEXER },
-#if defined(USE_SIM_VIDEO)
     { "VIDEO",          &vid_show,                  0, HLP_SHOW_VIDEO },
-#endif
     { "CLOCKS",         &sim_show_timers,           0, HLP_SHOW_CLOCKS },
     { "SEND",           &sim_show_send,             0, HLP_SHOW_SEND },
     { "EXPECT",         &sim_show_expect,           0, HLP_SHOW_EXPECT },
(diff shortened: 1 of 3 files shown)


larsbrinkhoff:
This is now a crude proof-of-concept pull request.  It's just barely enough to make the TT2500 simulator display two windows.  I'm posting this for comments and to check whether this is roughly on the right track, or if it's headed in the wrong direction.

The main gist of it is that the global window state is captures in a VID_DISPLAY struct.  A new API `vid_set_current` sets the index into an array of those.  Essentially, a simulator with multiple windows has to call this API before it calls any other sim_video API to indicate which window it's currently adressing.  This may sound awkward, and I welcome other suggestions.

On the event side, care must be taken to check which window an even comes from.  This hasn't been fully fleshed out yet since the TT2500 only has one keyboard and no mouse.

2500-dual

larsbrinkhoff:
I updated this to also handle events.

Tested with the TT2500 simulator and by playing GT40 Lunar Lander.

markpizz:
> I'd like to get this to a good state for merging.  Of course there's no urgency, so if other things have higher priority I will wait.  My pull request seems to be fully working, but there's the matter of extending the sim_video API.  (I haven't touched the vector display library.  It still works as one of the sim_video displays, but just one.)

As it turns out, I was just looking at this pull request yesterday and was drafting my comments and other feedback.

> My prototype has two added function calls:
> 
> * vid_set_current (int n) - Select display index n.  Future sim_video
  API calls will apply to display n.
> * vid_close_all () - Close all displays.

> The global vid_active variable is changed from a t_bool to an int, indicating how many active displays there are.  The event structs have added a DEVICE * to indicate from which display the even was generated.
> (Maybe change to UNIT *, see below.)

So far, so good, but let's keep away from entangling the video aspect and the simh UNIT structure.  Most specifically, due to the fact that there may be several independent DEVICEs and their related UNITs involved here (Video, Keyboard, and Mouse, Joystick in some cases, etc.).


> I suggest there could at the very least be a new function to dynamically allocate a new display and return an index (or pointer) for use with vid_set_current.  The current vid_open expects a DEVICE * for the display.  If there are many displays, it's awkward to make one DEVICE for each.  Maybe there could be a way to use a UNIT instead?
> 
t_stat vid_open_new (int *index, UNIT *unit, const char *title, uint32 width, uint32 height, int flags);

Specifically, what an API like this returns could be a DISPLAY* or a window number.  I don't think that matters too much.

Comments on your PR:

"More than one" in your current test case is merely 2.  Certainly the jump from 1 to 2 is most significant, but while you're here, let's try and make this more general and either make it dynamic, or make it static, but larger (larger is less desirable than dynamic).

So, with more than one window, I would have preferred that "legacy" window DEVICEs in all existing simulators use whatever vid_XXX APIs they have always used, with each of those APIs merely be some sort of wrapper that then invokes the newer implementation passing the "initial or default" window context structure (vid_close_all() is fine).  Meanwhile, newer window DEVICEs that support multiple windows should use "extended" vid_XXX APIs which provide some type of reference to the window context that they are accessing.  This approach would seem to avoid the rather continuous need for all sorts of things calling vid_set_current() all the time (which vid_set_curren then be un-necessary), Otherwise, please explain the thinking behind this...

The vid_screenshot() API should, if more than one window is active, screenshot all active windows and use the filename+window# as the desired filename for the set of screenshot files, otherwise in the single window case, behave as it did before.

Please pickup commit: 01801e01bcbe3517abf0 from the lars/dual branch at https://github.com/markpizz/simh and include it in your next revision or I can commit those minor changes to the master branch and you'll then pick up the pieces...  Let me know.

larsbrinkhoff:
Thanks for the feedback.  The current state is very much only a proof of concept, and I'm wide open to changes in both the implementation and the APIs.  Basically, whatever you think is best for SIMH.

> let's keep away from entangling the video aspect and the simh UNIT structure.

Ok.

> "More than one" in your current test case is merely 2.

That's just the prototype.  I imagine new displays should malloc their data and link into a list or other suitable data structure.

> each of those APIs merely be some sort of wrapper that then invokes the newer implementation passing the "initial or default" window context structure

Ok, I will then provide a new set of APIs that extends the old ones by taking a "display" (probably pointer) as input.  And `vid_set_current` will go away.

> The vid_screenshot() API should...

Ok.

> Please pickup commit: 01801e0

Ok, I will do that later today.

larsbrinkhoff:
I believe I have addresses all your points above.

- There are new APIs that parallel the old but adds a display pointer parameter.
- The display structs are now allocated dynamically and linked into a list.
- Taking a screenshot writes numbered fles.
- Your commit is included.

larsbrinkhoff:
If I may, I'd like to add another sim_video extension to this pull request.  It's to allow use of the alpha channel.  It's a separate commit.

markpizz:
> If I may, I'd like to add another sim_video extension to this pull request. It's to allow use of the alpha channel. It's a separate commit.

You are out of my domain of expertise here.  Please explain how this would be relevant in the context of a simulated graphical device or otherwise be relevant to simh.

Thanks.

larsbrinkhoff:
> Please explain how this would be relevant in the context of a simulated graphical device

My immediate use is for simulating a random-scan point-plotting CRT.  It's much like the PDP-1 Type 30 display supported by the display library, except it's a color TV tube.  So the points can be any color.  This particular device had a rather large dot size that could overlap with neighbouring dots.

I draw individual dots using a Gaussian distribution in a small square, like 7x7 pixels.  If two dots are drawn close, or even superimposed, the alpha channel will ensure additive color blending.  I first looked into using the display library, but it doesn't support this use case and only handles dual-phosphor tubes.

Other uses are possible, like antialiased vector lines.

larsbrinkhoff:
Here's a screencast of another simulation I made using an OpenGL shader:  
https://www.youtube.com/watch?v=3WH4yAwaU6I

Compare against historical film footage off the CRT:

markpizz:
I somehow thought I had passed along an explicit request to change the sim_video API names from XXX_display to XXX_window so as to not be so in bed with your primary use case in the display/ directory.

Please fix these.  Thanks.

larsbrinkhoff:
> I somehow thought I had passed along an explicit request to change the sim_video API names from XXX_display to XXX_window

Ok, fixed.

On second thought, I dropped the "alpha channel" commit to work on that separately.

markpizz:
The alpha channel looked simple enough.  Feel free to add it.  It doesn't break anything.

In the process of testing before merge, I notice these minor warnings:
```
1>sim_video.c
1>..\sim_video.c(386) : warning C4101: 'i' : unreferenced local variable
1>..\sim_video.c(843) : warning C4098: 'vid_draw' : 'void' function returning a value
1>..\sim_video.c(959) : warning C4098: 'vid_refresh' : 'void' function returning a value
```

larsbrinkhoff:
Thanks, those three warnings are fixed now.

I have compiled with gcc, g++, and clang++ to catch as many warnings as possible.  But not many are enabled.  Is there a small subset of useful warnings that could be turned on?  Maybe just `-Wunused-variable`?

Regarding the alpha channel, I found there are a few more changes needed to get the right effect.  I'll let the dust settle here first, and then work on a new pull request.

PR #959, opened 2020-11-17
AltairZ80: set sim_clock_precalibrate_commands and sim_vm_initial_ips

deltecent:
For @markpizz and @psco to consider.

### DIFF
--- a/AltairZ80/altairz80_cpu.c
+++ b/AltairZ80/altairz80_cpu.c
@@ -6180,12 +6180,28 @@ static t_stat sim_instr_mmu (void) {
     return reason;
 }
 
+/*
+ * This sequence of instructions is a mix that mimics
+ * a resonable instruction set that is a close estimate
+ * to the normal calibrated result.
+ */
+
+static const char *cpu_clock_precalibrate_commands[] = {
+    "-m 100 IN 08H",
+    "-m 102 ANI 01H",
+    "-m 104 XRA A",
+    "-m 105 JMP 0100H",
+    "PC 100",
+    NULL};
+
 /* reset routine */
 
 static t_stat cpu_reset(DEVICE *dptr) {
     int32 i;
     if (sim_vm_is_subroutine_call == NULL) { /* First time reset? */
         sim_vm_is_subroutine_call = cpu_is_pc_a_subroutine_call;
+        sim_clock_precalibrate_commands = cpu_clock_precalibrate_commands;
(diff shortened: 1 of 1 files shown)


psco:
This should be rejected as the "reasonable" instruction mix is in fact a busy wait loop on an uninitialised disk device.

deltecent:
My understanding is the instruction mix isn't really supposed to do anything useful. It is to be 3-4 instructions that exercise various paths through the VM to give SIMH a general starting point for timing. If there is a problem banging on port 8, then we can probably use FF.

>The simulator framework provides a means of pre-calibrating the instruction execution rate.  A sequence of three to four instructions that run in a tight loop can be used at simulator startup time to compute the execution rate.

markpizz:
The goal of the pre-calibrate exercise is to come up with a reasonable approximation of a  somewhat "normal" instruction mix that actually runs at a similar rate to how real instructions do when the simulator is doing something useful. 

I can't personally tell if the results, in this case, are true.  However, other simulators tend to get much faster instruction rates when the mix they test with doesn't include some percentage of non computational (i.e. instructions which reference some aspect of other pieces of the simulator which support I/O activities).  

Much faster pre-calibration is not better.

psco:
One problem is that if you do I/O (i.e. IN or OUT instructions, any port btw.) you are invoking parts of the simulator which are for all we know in basically an undefined state. The other problem is that having one out of four instructions an I/O instruction is hardly what is a "reasonable approximation of a somewhat "normal" instruction mix".

deltecent:
@psco it is probably best that you come up with an instruction mix, assuming you want to add pre-calibration to AltairZ80.

markpizz:
The goal is to get  a number which is similar to the numbers observed when doing something useful.  Surely the device(s) may be in an indeterminate state, but that state will be the same each time the simulator is started so some consistent behaviors/timings can be measured.

psco:
Honestly I am not keen at all to add yet another piece of timer related complexity with unclear benefit and a high potential of more trouble down the road. Furthermore I see only few simulators implementing this which is not encouraging either.

deltecent:
As I may be the only person that uses AltairZ80 in a way where timing is important, that course of action seems reasonable. I'll maintain the timing changes in my own fork. What I learned through this process is that the internal workings of the timer functions are mystical voodoo magic, but they seem to work when used appropriately.

PR #962, opened 2020-11-19
SCP: Adds VM-specific stopped message handler

deltecent:
Adds the ability for a VM to define its own stopped message for all
codes, including SCPE errors. For example, on the AltairZ80 simulator,
a custom stop message handler could also display CPU flags and registers:

Simulation stopped C0Z0M0E0I0 A=61 B=1108 D=B818 H=D398 S=BC33 P=CC71 IN 0Ah
sim> s

Step expired C0Z0M0E0I0 A=11 B=1108 D=B818 H=D398 S=BC33 P=CC73 MOV M,A

sim_vm_fprint_stopped_gen = &my_vm_stopped_gen;

t_bool my_vm_stopped_gen (FILE *st, t_stat v, REG *pc, DEVICE *dptr) {
    /* display stopped message */
    fprintf(st, "My VM stopped!");

    return TRUE;  /* display SCP stopped message */

    return FALSE; /* do not display SCP stopped message */
}

### DIFF
--- a/scp.c
+++ b/scp.c
@@ -451,6 +451,7 @@ void (*sim_vm_fprint_addr) (FILE *st, DEVICE *dptr, t_addr addr) = NULL;
 t_addr (*sim_vm_parse_addr) (DEVICE *dptr, CONST char *cptr, CONST char **tptr) = NULL;
 t_value (*sim_vm_pc_value) (void) = NULL;
 t_bool (*sim_vm_is_subroutine_call) (t_addr **ret_addrs) = NULL;
+t_bool (*sim_vm_fprint_stopped_gen) (FILE *st, t_stat v, REG *pc, DEVICE *dptr) = NULL;
 t_bool (*sim_vm_fprint_stopped) (FILE *st, t_stat reason) = NULL;
 const char *sim_vm_release = NULL;
 const char *sim_vm_release_message = NULL;
@@ -8909,6 +8910,10 @@ int32 i;
 t_stat r = 0;
 t_addr k;
 t_value pcval;
+
+if ((sim_vm_fprint_stopped_gen != NULL) &&              /* if a VM-specific handler is defined */
+    (!sim_vm_fprint_stopped_gen (st, v, pc, dptr)))     /*   call it; if it returned FALSE, */
+    return;                                             /*     we're done */
 
 fputc ('\n', st);                                       /* start on a new line */
 
--- a/scp.h
+++ b/scp.h
@@ -445,6 +445,7 @@ extern CTAB *sim_vm_cmd;
 extern void (*sim_vm_sprint_addr) (char *buf, DEVICE *dptr, t_addr addr);
 extern void (*sim_vm_fprint_addr) (FILE *st, DEVICE *dptr, t_addr addr);
 extern t_addr (*sim_vm_parse_addr) (DEVICE *dptr, CONST char *cptr, CONST char **tptr);
+extern t_bool (*sim_vm_fprint_stopped_gen) (FILE *st, t_stat v, REG *pc, DEVICE *dptr);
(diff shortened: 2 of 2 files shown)


markpizz:
Without any additional changes to SCP, there are already specific message format related and other features which are available.

All RUN related commands (RUN, BOOT, GO, CONTINUE, STEP, NEXT, etc.) invoke fprint_stopped() as a result of this being the message format routine for the run_cmd dispatch.  This routine already has an optional simulator specific provided routine to, in a simulator specific way, to format simulator specific stop return codes via the sim_vm_fprint_stopped function pointer.  This indeed is only dispatched to for simulator specific sim_instr() exit/stop codes (i.e. not for normal SCPE_xx codes).

Meanwhile, you could:
```
sim> SET ON
sim> ON STEP EXAMINE CPU STATE
```
and have all the CPU REGisters displayed.  In their "natural" format (where "natural" is the precise way they are described in the each REGister's definition.  Such a REGister definition can already display both a complete register value as well as any internal bit fields that the register may contain (again with a proper REGister description that describes the interesting bit fields).
For example, the PDP11 simulator  might want to:
```
sim> SET ON
sim> ON STEP EXAMINE CPU PC,PSW,R0,R1,R2,R3,R4,R5
sim> STEP
Step expired, PC: 000100 (INC 120)
PC:     000100
PSW:    000000  CM=K PM=K RS0 FPD0 IPL=0 TBIT0 N0 Z0 V0 C0
R0:     000000
R1:     000000
R2:     000000
R3:     000000
R4:     000000
R5:     000000
sim>
```

This leaves the user (vs the simulator writer) the flexibility of precisely specifying what they're interested on this particular SCPE_xx stop code, but also potentially different setup for other stop codes they may desire.  All without changing simulator code.

rcornwell:
This could be somewhat useful. I added the stop function to my IBM360 simulator to display the PSW. What is interesting is that when you add the sim_vm_fprint_stopped it is only called for user stop codes. I think it should be called for any stop code.

Also a fprint_register function could be useful.

markpizz:
> This could be somewhat useful. I added the stop function to my IBM360 simulator to display the PSW. What is interesting is that when you add the sim_vm_fprint_stopped it is only called for user stop codes. I think it should be called for any stop code.


As I mentioned, there already is a sim_vm_fprint_stopped function pointer which can optionally be supplied by a simulator to format simulator specific stop code information which are presumably unique to each simulator.

> Also a fprint_register function could be useful.

After you add BITFIELD REGister declarations where appropriate (as is demonstrated in the previous PDP11 PSW example), to some of your simulators, please come back and describe the problem you need to solve.

rcornwell:
BITFIELD seems very useful, however it will not work for the IBM360 since there are 2 formats for the PSW based on the mode the CPU is in.

markpizz:
> BITFIELD seems very useful, however it will not work for the IBM360 since there are 2 formats for the PSW based on the mode the CPU is in.

Nothing says that you can't have multiple REGister definitions, pointing at exactly the same spot in memory each with different BITFIELDS defined.  It seems that your IBM360 doesn't define PSW as a simh REGister anyway...  Should it?

rcornwell:
The format of the PSW is controlled by the type of the processor and the mode it is in. BC or EC/67 EC/370. The user just want do do Examine/Deposit PSW, not have to know which version to use. This is why a call the examine/deposit a register would be useful. A flag in the REG entry could signal the need to call either of these functions. It would then pass the REG * and value or return a value. This would no break any existing simulators, and simplify some of them. For simulators where there was a bunch of register bits stored in different variables BITFIELD would work.

Also if sim_vm_fprint_stopped is defined this should be called regardless of the error code. If the user does not wish to handle it, the code can do the current default of printing the PC register.

deltecent:
>This [PR] would not break any existing simulators, and simplify some of them.

I agree.

PR #965, opened 2020-11-24
New simulator: Turtle Terminal 2500, from General Turtle Inc.

larsbrinkhoff:
Custom TTL design by Marvin Minsky.  There are two displays: one raster scan for bitmapped characters, and another random scan for vector graphics.  There is also a keyboard, and a UART for talking to a host computer.

The processor is intended for microcoding rather than a general-purpose computer.  There is no interrupt facility, so the software must check frequently for hardware wanting attention.  Some pipeline hazards are exposed.  The microcode was written by Danny Hillis, and implements a Datapoint terminal emulator and Logo turtle graphics.


markpizz:
There are issues with this code.

First, the choice of a variable name CHAR (as an array) which in some places collides with other names that may be defined system wide.

Beyond that, we've got:

1>..\tt2500\tt2500_cpu.c(218) : warning C4033: 'mem_read' must return a value
1>..\tt2500\tt2500_cpu.c(226) : error C2143: syntax error : missing ';' before 'type'
1>..\tt2500\tt2500_cpu.c(227) : error C2181: illegal else without matching if
1>..\tt2500\tt2500_cpu.c(383) : warning C4098: 'cpu_bus' : 'void' function returning a value
1>..\tt2500\tt2500_cpu.c(388) : warning C4098: 'cpu_bus' : 'void' function returning a value
1>..\tt2500\tt2500_cpu.c(389) : warning C4098: 'cpu_bus' : 'void' function returning a value
1>..\tt2500\tt2500_cpu.c(390) : warning C4098: 'cpu_bus' : 'void' function returning a value

There may be other problems concealed by the earlier errors with the CHAR array

larsbrinkhoff:
Thanks, I didn't know about CHAR being reserved.  The problems should be fixed now.

markpizz:
```
1>tt2500_dpy.c
1>..\tt2500\tt2500_dpy.c(213) : warning C4244: '=' : conversion from 'uint16' to 'uint8', possible loss of data
1>tt2500_cpu.c
1>..\tt2500\tt2500_cpu.c(226) : warning C4244: '=' : conversion from 'uint16' to 'uint8', possible loss of data
```

markpizz:
Looking through the tt2500 and imlac simulators you've got conditional compiles looking at HAVE_LIBSDL which then disables some sim_video code references.

The presumption is:  In general simulator specific code generally shouldn't have any platform specific references (specifically HAVE_LIBSDL in this case).  At most, such references "should" only be used to potentially provide the default enabled or disabled state for some DEVICEs.
The simulator code's references to the video functionality (in this case), when video isn't available, should end up being references to the stub routines for the various APIs which are available in the "not available" path in sim_video.  There should be stubs for the new XXX_window APIs that sim_video provides, but that's not the real problem here.  Somehow the pdp11 simulator doesn't have any references to HAVE_LIBSDL, and yet it builds both with and without libsdl being available.  All the VAX simulators that have different video devices, only have references to USE_SIM_VIDEO && HAVE_LIBSDL to control enabling the respective video DEVICE's and not to influence calling various APIs.  The pdp11 uses both sim_video and display/display.c.

I'm trying to figure out how the pdp11 and vax simulators work out but the the tt2500 and imlac don't...

Please look into adding stubs to the newer sim_video API's.

larsbrinkhoff:
I added the stubs in #967.

larsbrinkhoff:
I updated the TT2500 simulator to use the sim_video stubs rather than #ifdefs.  The #ifdefs for the display library are still in place until I know what to do with them.

I see the PDP11 simulator has #ifdef USE_DISPLAY everywhere the display library is referenced.  Is that the right thing to do?

markpizz:
> > I'm trying to figure out how the pdp11 and vax simulators work out but the the tt2500 and imlac don't...

I was confused by my own doing starting from the Visual Studio projects (which always have libsdl provided by the windows-build dependent parallel repo) and thus was looking in vt11.c which always was calling display functions.  In the non libsdl case (as spelled out in the makefile) vt11.c doesn't get included in the pdp11 build.

> I see the PDP11 simulator has #ifdef USE_DISPLAY everywhere the display library is referenced. Is that the right thing to do?

Given the structure of the makefile it certainly seems so.  Go with that.  Thanks.

------

Your travis,yml is building the tt2500 twice.

------

I picked up your lars/unused commit.

------

I need more sleep.  I'll look at your lars/tt2500 stuff later.

larsbrinkhoff:
- Updated #ifdefs to reference USE_DISPLAY.
- Fixed duplication in .travis.yml.
- By all means, take your time.

markpizz:
The first non comment line of  tt2500_test.ini should be:
```
cd %~p0
```
This changes the directory to the location of tt2500_test.ini

Then the load command should merely be
```
load test.ascii
```
I would expect that the RESET followed by GO 100 should be replicable by RUN 100, but as you have mentioned this somehow isn't working.  I'm looking at that problem.

markpizz:
The reason for changing the directory is due to the fact that the default directory during the Visual Studio build is not the top directory in the repo.

larsbrinkhoff:
Thanks, updated.

markpizz:
The RESET + GO nnn being equivalent to RUN nnn is now fixed.

The problem, in your case, was the fact that your CPU DEVICE reset routine was reasonably setting the PC to 0, while the original scp run_cmd implementation set the PC value to nnn before calling the reset routine.  It now sets the PC value after invoking the reset routine.

larsbrinkhoff:
Thanks!  Updated.

markpizz:
Your hand edited Visual Studio solution was broken.  Use this one.

[simh-sln.zip](https://github.com/simh/simh/files/5625121/simh-sln.zip)

markpizz:
Fix the commit time and I'll merge it!

larsbrinkhoff:
Commit time updated!

larsbrinkhoff:
I made another update, but I need to test it more before merging.

markpizz:
I didn't see your most recent comment before I merged.  Oh, well, fix as needed later... 😁

PR #967, opened 2020-11-25
VIDEO, IMLAC: Add new sim_video stubs.  Use them in Imlac simulator.

larsbrinkhoff:
As per https://github.com/simh/simh/pull/965#issuecomment-733172291.

This cleans up *some* uses of #ifdef HAVE_LIBSDL in the Imlac simulator.  The sim_video component offers stubs when SDL isn't available, but the display library doesn't and isn''t even linked in by the makefile.

### DIFF
--- a/sim_video.h
+++ b/sim_video.h
@@ -217,7 +217,6 @@ uint32 vid_map_rgb_window (VID_DISPLAY *vptr, uint8 r, uint8 g, uint8 b);
 void vid_draw_window (VID_DISPLAY *vptr, int32 x, int32 y, int32 w, int32 h, uint32 *buf);
 void vid_refresh_window (VID_DISPLAY *vptr);
 t_stat vid_set_cursor_window (VID_DISPLAY *vptr, t_bool visible, uint32 width, uint32 height, uint8 *data, uint8 *mask, uint32 hot_x, uint32 hot_y);
-t_stat vid_show_video_window (VID_DISPLAY *vptr, FILE* st, UNIT* uptr, int32 val, CONST void* desc);
 t_bool vid_is_fullscreen_window (VID_DISPLAY *vptr);
 t_stat vid_set_fullscreen_window (VID_DISPLAY *vptr, t_bool flag);
 void vid_set_cursor_position_window (VID_DISPLAY *vptr, int32 x, int32 y);        /* cursor position (set by calling code) */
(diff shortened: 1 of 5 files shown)


sim_video.h
```diff
@@ -217,7 +217,6 @@ uint32 vid_map_rgb_window (VID_DISPLAY *vptr, uint8 r, uint8 g, uint8 b);
 void vid_draw_window (VID_DISPLAY *vptr, int32 x, int32 y, int32 w, int32 h, uint32 *buf);
 void vid_refresh_window (VID_DISPLAY *vptr);
 t_stat vid_set_cursor_window (VID_DISPLAY *vptr, t_bool visible, uint32 width, uint32 height, uint8 *data, uint8 *mask, uint32 hot_x, uint32 hot_y);
-t_stat vid_show_video_window (VID_DISPLAY *vptr, FILE* st, UNIT* uptr, int32 val, CONST void* desc);
```
larsbrinkhoff, line 220: Removed because there is no such function.

markpizz:
In case you hadn't noticed, due to all the other Travis and AppVeyor messages, this commit broke the imlac Windows build.

The specific change that actually caused this is the fact that the CRT DEVICE is now ENABLED when libsdl is available.

As it turns out, the failure is NOT related to any of the recent sim_video changes.  Setting the CRT DEVICE ENABLED prior to the recent changes in sim_video() causes the same problem.

The problem is indicative of an infinite loop producing this message:
```
vid_thread(): Unexpected user event code: 8
```

The problem is platform specific due to the fact that on some platforms the events are processed on either the main process thread or on a separate thread depending platform requirements.

In any case, the CRT DEVICE is now presenting a window very early in SCP's startup sequence BEFORE all of SCP has been properly initialized.  The build invokes the simulator with "RegisterSanityCheck" that is designed to walk all the REGister definitions in the simulator to assure that they have reasonable definitions which can't be done by a compiler.  This check then immediately exits the simulator indicating success or failure of that sanity check for the build.  The problem here is due to the immediate exit of SCP's main routine without going through all the cleanup steps that always occur when the simulator has been fully initialized.  I've changed scp to unwind everything on the way out in this case and the build problem is now solved.

Meanwhile, maybe it isn't a good idea for windows to pop up windows this early in the simulator's execution.  It pops up due to the "reset_all_p()" call that exists to "power up" all the simulator's DEVICES.  DEVICEs can distinguish this call to their reset routine by the presence of the SWMASK('P') in sim_switches.  DEVICE reset routines will be called again later as a consequence of one of: RESET, RUN or BOOT commands.

larsbrinkhoff:
Sorry, I didn't notice.  I will hurry to fix it.

PR #968, opened 2020-11-25
I650, IBM1130, PDP11, PDP18b, VAX, Sigma, CARD: Fix set but unused variables.

larsbrinkhoff:
And some other warnings.

All simulators were compiled with gcc, g++, and clang++ with -Wunused-variable enabled.  All warnings except the dummy_declaration in VAX files were fixed.

### DIFF
--- a/I650/i650_dsk.c
+++ b/I650/i650_dsk.c
@@ -453,8 +453,6 @@ t_stat dsk_reset(DEVICE * dptr)
 
 t_stat dsk_attach(UNIT * uptr, CONST char *file)
 {
-    DEVICE             *dptr = find_dev_from_unit(uptr);
-    int                 unit = (uptr - dptr->units);
     t_stat              r;
     int                    flen;
 
--- a/Ibm1130/ibm1130_t2741.c
+++ b/Ibm1130/ibm1130_t2741.c
@@ -5,12 +5,8 @@
 
 #include "ibm1130_defs.h"
 #include "sim_sock.h"
-#include "sim_tmxr.h"
 
 #define DEBUG_T2741
-
-static TMLN t2741_ldsc = { 0 };                         /* line descr for telnet attachment */
-static TMXR t2741_tmxr = { 1, 0, 0, &t2741_ldsc };      /* line mux for telnet attachment */
 
 #define T2741_DSW_TRANSMIT_NOT_READY            0x4000
 #define T2741_DSW_READ_RESPONSE                 0x1000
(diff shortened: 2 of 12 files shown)


rcornwell:
Do not remove ascii_to_hol_ebcdic this may be used by in future if I decide to add EBCDIC translation to sim_card.

larsbrinkhoff:
I reckoned it might be useful in the future, so I just `#if 0`'ed it.

My thinking is that this commit makes it possible to add `-Wunused-variable` to the makefile to catch more mistakes in the future.  But it's up to @markpizz whether he wants to do that.  If so, I could fix some other warnings too.

psco:
I fail to see why type casts such as
```
char *fmt;
...
fmt = "7E2"; // old
fmt = (char *)"7E2"; // new
```
are useful. I do not want this in the AltairZ80 files.

@markpizz Please reject the changes for the AltairZ80.

@larsbrinkhoff Please reach out to me first before proposing any changes. I maintain the AltairZ80 and I'm happy to perform any changes which have a real benefit.

larsbrinkhoff:
@psco, no problem I will update the commit.

The reason is that C++ is picky about some things.  A `void *` will not cast automatically to other pointers like in C.  A string constant isn't a plain `char *`.

markpizz:
Don't bother updating it.  I will take these changes in pieces as appropriate for the different code

markpizz:
Additionally, we really don't have to get rid of every theoretical warning produced by -Wall.  That is why many/most of these have been left in place already.

larsbrinkhoff:
> we really don't have to get rid of every theoretical warning produced by -Wall.

Undrstood.  I only picked -Wunused-variable because apparently the Windows compilers warn about them, but gcc and clang doesn't.

markpizz:
Actually, the windows compiler only flags variables that are declared but never used for anything.  

Merely setting a value as part of the declaration "counts" as being "used"  by that compiler even if it is subsequently never used again.

psco:
Actually I am a big fan getting rid of unused variables (and functions for that matter) as they are confusing and often point to a deeper issue. But I'm pretty sure you did not find any in AltairZ80 (otherwise I'm very interested to know).

markpizz:
> @markpizz Please reject the changes for the AltairZ80.

As I said, I was going to bring the AltairZ80 stuff directly with you.  Lars was a little over aggressive, but well meaning.

Historically, I occasionally compile the simulators with as many compilers as possible (across platforms) to get as much natural feedback.  This includes compiling as C++.  I haven't done that in quite a while, nor have we had a Coverity scan which might have flagged some of the same issues.

In any case, various things in the code that Patrick has contributed recently could use a C++ cleanup.
This mostly means changing various "char *" to "const char *" reasonably casting void * into the type we're expecting.
Look at the output of "make altairz80 GCC=clang++" and see if find anything that should be cleaned up...

Thanks.

psco:
Perfect - I'll submit a PR for the cleaning up.

larsbrinkhoff:
I rebased this.  But feel free to close it or cherry-pick bits and pieces as appropriate.

markpizz:
Cherry picked the relevant changes.

PR #972, opened 2020-11-28
AltairZ80: New iCOM device, PTP port command, code cleanup

psco:


markpizz:
One issue noted in Windows build:

2>s100_icom.c
2>..\AltairZ80\s100_icom.c(1172) : warning C4244: 'function' : conversion from 'uint16' to 'uint8', possible loss of data

also s100_icom.c also has some tabs.

PR #975, opened 2020-11-29
IMLAC: Sample program: SSV 22.

larsbrinkhoff:
SSV ("Scroll Saver") was the Imlac console program and "terminal emulator" at MIT's Dynamic Modeling group.  It was written mainly by @dlebling618.  This copy is from Purdue University, courtesy of @tuban: http://www.ubanproductions.com/imlac_sw.html

SSV is required to run Maze; the game expects to find SSV's font data in core.  The game loading process includes jumping to location 40, which is why the sample configuration enables the "Special TTY" ROM bootstrap.


markpizz:
I notice in your ssv.simh file you have:
```
# Reset and run.
reset
go
```
FYI: There is a simpler way to say this:
```
run
```

As I mentioned with you adjusted the reset logic in the imlac simulator:

> DEVICE reset routines will be called again later as a consequence of one of: RESET, RUN or BOOT commands.

Why is this stuff, by itself, being added to the simh repo?  If it is part of something bigger, shouldn't that also be here?

larsbrinkhoff:
I first tried `run`, but it didn't work the same as `reset` plus `go`.  I haven't investigated why.

Since there was no imlac/tests directory at all, I submitted this as a useful test and also a way to use the simulator as a terminal.  The bigger picture includes ITS running on a PDP-10 simulator, but that's not appropriate to check into the SIMH repository.

PR #978, opened 2020-12-11
CI: Build using GitHub Actions.

larsbrinkhoff:
Continuous integration build using GitHub Actions.  See https://github.com/features/actions

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,35 @@
+name: Build
+
+on: [pull_request]
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-10.15, ubuntu-18.04]
+        simulators:
+          # These are supposed to match ALL in makefile.
+          # Each job builds 15 simulators.
+          - pdp1 pdp4 pdp6 pdp7 pdp8 pdp9 pdp10 pdp10-ka pdp10-ki pdp10-kl pdp11 pdp15 vax microvax3900 microvax1 rtvax1000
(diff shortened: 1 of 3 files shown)


larsbrinkhoff:
Here's what it should look like after the branch has been merged:  
https://github.com/larsbrinkhoff/simh/actions

markpizz:
For the uninformed (me), please explain the net result of this.  What is the logical change?  Will Travis go away?  Will the osx builds be less likely to fail for "environmental" reasons rather than actual issues in the simh codebase?

larsbrinkhoff:
I have only been testing this for a few hours, so I don't know if the GitHub Actions will be more stable than Travis CI.  The the builds seem good so far, though.  Certainly much faster.

Rumor has it that Travis CI will reduce their service: https://twitter.com/james_hilliard/status/1336081776691843072
I think I have seen jobs staying queued up for longer lately.

So at this point, the net result is a redundant second source of CI builds.  In the future, it could be an exit path away from Travis CI.

markpizz:
To me, it seems somewhat wasteful to have some system(s) spend what seems like a ton of cycles doing all of the build work on every "push" to the repo.  Having BOTH GitHub and Travis do it would then be doubly so.  

If this is the new norm, then who am I to say it shouldn't be done.  Is there some way to limit these activities to at most 'n' times per some time interval?

markpizz:
> I remember your concern about wasted cycles, which is why I set on: [pull_request] I don't know what it means exactly, but roughly that builds are only started when a pull request is opened. Commits pushed to regular branches will not be built.

That seems good for pull requests since they really don't come at a crazy pace.  However, Travis stuff seems to run every time I push one (or more) commits to the repo.  This inspires me to hold back some pushes, to avoid wasting cycles after a spelling error correction, but that then leaves the HEAD of the master branch on my local system ahead of GitHub and thus potentially staging sequencing issues with newly arriving pull requests...  I merely get to deal with it...

markpizz:
How about we run with both GitHub Actions and Travis for a month and then take a measure of the advantage of GitHub Actions.  If it does well, then we retire the Travis stuff...

larsbrinkhoff:
Yes, that sounds like a good plan.

PR #986, opened 2021-01-02
BESM6: Implemented punched card input and punched tape output

leobru:


leobru:
What's the use compiling with -std=gnu99 for Linux if Visual Studio requires a much older standard?

markpizz:
Leo,

Somehow besm6_vu.c ended up with a couple of tab characters,  Please fix and force push again.

Thanks.

- Mark

markpizz:
I'm not sure when or who supplied the -std=gnu99, but it doesn't hurt, right?

leobru:
> I'm not sure when or who supplied the -std=gnu99, but it doesn't hurt, right?

It's an inconvenience. For example, when a C99 feature (mid-block variable declaration) was used, it compiled successfully in Linux, and in MacOS during the pull request pre-check, but failed to compile in Visual Studio. Having a consistent language standard across platforms will make the pull request process more efficient.

markpizz:
I hear your problem.  

I don't see an obvious change to the -std=gnu99 setting that will catch the problem you observed, but I'm open to suggestions...

That's one of the reasons I build each locally before accepting changes.

WILLIAMCORCORAN:
Hello,  can I get a status as to when my order will ship?  Thank you!


Truly,

Bill Corcoran


On Jan 2, 2021, at 10:30 PM, ReActiveMicro.com <sales@reactivemicro.com> wrote:

﻿

I'm not sure when or who supplied the -std=gnu99, but it doesn't hurt, right?

It's an inconvenience. For example, when a C99 feature (mid-block variable declaration) was used, it compiled successfully in Linux, and in MacOS during the pull request pre-check, but failed to compile in Visual Studio. Having a consistent language standard across platforms will make the pull request process more efficient.

PR #987, opened 2021-01-02
Enable warnings in build

promovicz:
This patch enables -Wall on all makefile builds so that people can see
how bad it has gotten.


promovicz:
Sorry, it was not my intention to push this to mainline simh.

PR #989, opened 2021-01-03
allow .git to be a file or directory to handle submodules

### LINKED ISSUE #988: makefile causes error when repo is included as a submodule
When simh is a submodule of another repo, the .git/ directory is not a directory, but a file. See https://git-scm.com/docs/gitrepository-layout.

The makefile has a _hack_ that attempts to workaround this, but in an unnecessarily complicated and very broken way. It attempts to see if the parent directory is a git repo with `cd .. ; git rev-parse --git-dir` and then uses `git submodule` commands to attempt to find the current commit. There are a number of problems with this:

- the repo should be self-contained and not make assumptions about other directories outside it.
- what if the submodule is "mounted" two layers in?
- what if the .git link is not part of a submodule setup? (which is possible, but unlikely)

Furthermore, if the parent directory is another git repo, the assumption is made that it is a submodule situation, but that code is also flawed. If there are more than zero other sibling subrepos (concretely, for example simtools), the line `git submodule status | grep "$(notdir $(realpath .))" | awk '{ print $$1 }'` will return **multiple results**. This causes much carnage and general badness.

This is a serious problem because the build fails. It essentially makes the repo completely unbuildable when a submodule or any other time .git happens to not be a proper directory. There are lots of other edge cases where the .git directory is in the "normal" layout, so making these kinds of assumptions is dangerous.

The simple, immediate, fix is to change the git repo detection from `-d .git` to `-e .git`. This trivial fix works fine for the immediate problem because then the current directory is detected as a git repo and everything is "fine".

In the further future, it would be better to:
- detect whether the build is occuring in a .git repo more robustly. Like with `git rev-parse --git-dir` or maybe even a better way.
- create the .git-commit-id file with proper makefile rules and dependencies instead of `$(shell )` commands

anders-code:
fixes #988

### DIFF
--- a/makefile
+++ b/makefile
@@ -273,7 +273,7 @@ ifeq (${WIN32},)  #*nix Environments (&& cygwin)
       OS_CCDEFS += -Wno-deprecated
     endif
   endif
-  ifeq (git-repo,$(shell if ${TEST} -d ./.git; then echo git-repo; fi))
+  ifeq (git-repo,$(shell if ${TEST} -e ./.git; then echo git-repo; fi))
     GIT_PATH=$(strip $(shell which git))
     ifeq (,$(GIT_PATH))
       $(error building using a git repository, but git is not available)


markpizz:
Before looking  closely at this, the goal of the messy things that are being done (which certainly may not be being done in a reasonable way) is explicitly to determine the commit-id of simh that is being used and not the commit of any repo it may be contained in.

Folks have been known to either submodule, or worse yet, randomly copy simh pieces (all, most, or only some) to some other repo and then the results have no useful connection to the state of the repo (and the point in time) that what they're working with.  Then they come along and ask questions.

If they're going to do this, starting from a git archive of the simh repo would be a better solution since the archive has sim_rev.h populated with the the commit the archive was made from.

anders-code:
I totally understand and this patch results in exactly the behavior you want - the .git-commit-id file contains the git commit ID of simh not some random other repo.

The issue is that instead of using `git submodule` in the parent repo (which is unreliably detected) and using awk to parse out the simh git commit, just let the "I am a git repo" test use -e .git instead of -d .git so that the normal git commands work normally.

markpizz:
If we're going to mess with anything here, I want to actually test it.  Do you have a repo that references simh as a submodule in the way you want to use it that I can clone and test with?

anders-code:
Yes, I should be able to post something this evening.

anders-code:
please take a look a : https://github.com/anders-code/vax-unix
it's a work in progress
`make submodules`
`make -j8`

markpizz:
OK.  That change seems to work as you describe.

Three things:
1) Your commit is against an earlier state of the repo (i.e. the master branch already has another commit after yours), so "merging it" will produce a merge commit.  I strive to avoid that level of noise.
2) Your github profile doesn't have your last name...
3) The commit message doesn't follow the general convention we try to shoot for in this repo.  which is:

affected-simulator-or-code-section: 1-line summary of activity less than 80
<optional-blank-line-if-extended-commit>
extra-lines-wrapped-before-80

I can cherry pick this commit and list you as the author.

or you can base the change on the head of the master branch, changing the commit message along the way.

A good commit message for this would be:
```
makefile: allow .git to be a file or directory to handle submodules

Fixes #988
```

This is certainly a lot of  messing for a 1 character change to the makefile.  :-)

anders-code:
1. absolutely, I definitely prefer the rebase (or rebase and merge) workflow as well. Honestly I'm used to bitbucket and gitlabs and have not used github merging of PRs, but had assumed this was an option in the web too at merge time (if there weren't conflicts or other changes). Updated and rebased.

3. absolutely! I also use and prefer this style. What you see here is an abortive attempt to do the **one bit** code change with the stupid github online editor. It did not go well (I got some editor garbage pasted in the wrong place, can't see the wrap margins, etc.) vi is the way to go.

Yes, a simple change, but the amount of back-forth is more to do with process and is not incommensurate with the time it took to track down - the makefile was not the easiest thing to figure out. So if I can fix a tricky problem with a single bit flip, I believe that is the theoretical ideal situation!

Thank you for being so responsive; this is sadly not all universal among other projects. Yay!

markpizz:
No problem.  Reasonable and correct requests are always engaged.

Except for managing issues (and PRs which are counted as issues), I don't drive things from the web UI.

Would you please update your GitHub profile to contain your full name???

PR #997, opened 2021-01-17
AltairZ80: vfdhd: Fix unit detach.

hharte:
FYI, @psco


psco:
Looks good

markpizz:
Actually.  It could be improved since there are now return statements in a row (lines 340 & 342).  Only the first will be effective and the second will be  dead code...

psco:
You're absolutely right. And while we are at it, one could remove the variable `r` altogether and have
    `return detach_unit(uptr);  /* detach unit */`
as the last statement of `vfdhd_detach`.

hharte:
Thanks for the feedback, updated.

PR #1004, opened 2021-02-01
PDP11: New VT11 setting to enable/disable switches.

### LINKED ISSUE #1001: display: Typing on keyboard frobs switches.
When using the PDP11 simulator in its GT40 configuration, typing on the keyboard changes the contents of the SR switch register.  This is especially apparent when running MIT's Datapoint emulator software, because one of the switches toggles between italic and normal text.

Originally the simulator only used the keyboard to operate the switches.  When keyboard support was added, the original function remained effective in parallel with the new typing support.

larsbrinkhoff:
Use `SET VT KEYBOARD=SWITCHES` or `NOSWITCHES` to enable/disable toggling switches with the keyboard.  Default is `SWITCHES`.

Resolves #1001.

### DIFF
--- a/PDP11/pdp11_vt.c
+++ b/PDP11/pdp11_vt.c
@@ -81,6 +81,9 @@
  */
 #define CYCLE_US (MEMORY_CYCLE*(VT11_DELAY*2+1))
 
+#define UNIT_V_NOSPACEWAR  (UNIT_V_UF + 0)
+#define UNIT_NOSPACEWAR    (1 << UNIT_V_NOSPACEWAR)
+
 extern int32 int_vec[IPL_HLVL][32];
 
 t_stat vt_rd(int32 *data, int32 PA, int32 access);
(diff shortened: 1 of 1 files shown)


markpizz:
On Monday, February 1, 2021 at 2:41 AM, Mark Pizzolato wrote:
> A few things about this:
> 
> 1) The default should be the historic behavior
> 2) I'm thinking the option should be specifically admit the spacewar
> keybindings.  Simply mentioning SWITCHES may be meaningful to you, but it
> is less than clear to other folks.
>            SET VT KEYBOARD=SPACEWAR
>   and  SET VT KEYBOARD=NOSPACEWAR
> 3) The default MTAB behavior sets UNIT.flags, not DEVICE.flags, which is fine.
> You should define UNIT_SWITCHES instead of DEV_SWITCHES.  If you really
> want to use a DEVICE flag, then you'll need a specific MTAB set routine you
> should have a MTAB show routine.

Your updated change is better.

In vt_set_kb() you should use (MATCH_CMD(gbuf, "SWITCHES") == 0) instead of (strcasecmp(gbuf, "SWITCHES") == 0), likewise for "NOSWITCHES".  If neither matches, then you should report an error with return sim_messagef (SCPE_ARG, "Unexpected Keyboard setting: %s\n", gbuf);

In vt_show_kb() the string "KEYBOARD=" should be "keyboard=" to be consistent with the various other details of this device:

```
sim> set vt ena
sim> show vt
VT      crt=VR17, scale=2, hspace=normal
        vspace=normal, address=17772000-17772007, vector=300-314*, BR4
        KEYBOARD=SWITCHES
sim> set vt key=nosw
```
No error reported!
```
sim> show vt
VT      crt=VR17, scale=2, hspace=normal
        vspace=normal, address=17772000-17772007, vector=300-314*, BR4
        KEYBOARD=SWITCHES
sim> set vt key=noswitches
sim> show vt
VT      crt=VR17, scale=2, hspace=normal
        vspace=normal, address=17772000-17772007, vector=300-314*, BR4
        KEYBOARD=NOSWITCHES
sim> 
```
Still no mention of this existing for spacewar.  Why not?  The MTAB entry string "keyboard" is lacking useful detail.  It could be "Disable or Enable Spacewar switches".  Also, the MTAB string which currently says "KEYBOARD=SWITCHES" should mention the choices: "KEYBOARD={NOSWITCHES|SWITCHES}"

larsbrinkhoff:
Updated per comments.

PR #1008, opened 2021-02-15
AltairZ80: Add MDS-AD support for single-density

hharte:
The North Star MDS-AD disk controller supports both single- and
double-density operation.  While mixed-density is supported by the
controller, it is not supported by AltairZ80 due to a limitation of the
.nsi image file format.  The .nsi image file format requires all sectors
to be of the same density.


hharte:
FYI, @psco

psco:
Thanks Howard!

markpizz:
@psco, does Thanks mean you're OK with this?

psco:
Yes, this PR is fine.

PR #1011, opened 2021-02-23
M9312 implementation

friesga:
Hi Mark,

This is a pull request for implementation of the HITMISS register and the m9312 in the pdp-11. I tried to separate the pull requests but the m9312 implementation is dependent on the HITMISS changes, as they both effect pdp11_sys.c. The changes for the HITMISS register are in pdp11_cpu.c, pdp11_cpumod.h, pdp11_defs.h, pdp11_io.c, pdp11_rh.c en pdp11_sys.c. For the m9312 implementation I added two files: pdp11_m9312.h and pdp11m9312.c, and added a reference to m9312_dev in pdp11_sys.c. Feel free to ask if you have any questions!

Kind regards,

Jos


markpizz:
Hi Jos,

Glad you put all of it together so the full needs of HITMISS behavior have been determined.  Meanwhile, I'm looking first at the details of your resulting HITMISS changes.  I'm considering this pull request as a work in progress until we get things looking well all around.

My first goal here is to coalesce the HITMISS stuff so that I can pass it to back to Bob Supnik (the author of the PDP11 simulator) with minimal changes to the existing code for his review and advice.

My initial reactions to these changes are:
1) You may notice that the existing PDP11 codebase has very minimal specific references to CPU specific options scattered throughout the various simulator & DEVICE source modules (only one reference to a CPU model feature (OPT_RH11 exists in pdp11_rh)).  This suggests that the "if (CPUO (OPT_CACHE))" tests actually belong within the setHITMISS routine.
2) Not knowing precisely the specific behavior of the cache in these systems, it would seem that the actual call to setHITMISS() would belong directly in RdMemW(), WrMemW(), RdMemB(), and WrMemB().
These changes would then mean that pdp11_rh.c, pdp11_io.c and pdp11_sys.c would not need to be modified for proper HITMISS behavior.

Related to this, the setHITMISS routine would seem to belong in the pdp11_cpumod.c module since that is where CPU model specific features are implemented.  This would then remove the need for changes in pdp11_cpu.c.

Hmm... RdMemB, RdMemW, WrMemB and WrMemW are macros.  These macros can be changed to invoke new routines in pdp11_cpumod.c which then implement the similar logic as is already done for the UC15 variant.  This gets into where Bob would make the call about doing this.  The consequence here is that we're adding performance implications to essentially all simulated memory references.  My gut says that Bob would argue for accuracy and simplicity in implementation so the performance hit will be justified, and likely very hard to actually observe in running simulators.  From my quick observation this simplifies the HITMISS changes to involve pdp11_cpumod.c., pdp11_cpumod.h, the OPT_CACHE change and redefining RdMem* and WrMem* macros in pdp11_defs.h.

Lets see if you can do just the HITMISS logic in a separate branch in your repo.  Once that looks good, the M9312 stuff can derive from that HITMISS branch (which will ultimately be added to the master branch as a single commit squashing your various steps along the way).  

Please avoid committing changes to simh.sln since that file gets modified by the newer Visual Studio versions when they convert the VS2008 sln and proj files.  Committing that breaks things for VS2008.  We don't maintain other VS version details in the repo since they can all be derived dynamically from the VS2008 info.  "Visual Studio Projects/0ReadMe_Projects.txt" describes where you can find the installer for VS2008 which installs cleanly on all Windows environments since Vista.  Use VS2008 or your current version, but don't commit simh.sln.

Looking ahead to the M93212 details, it would seem useful to enrich the m93211_help routine to explain  details about the available device ROMs and provide specific examples of how to configure them.  Please, also explain which systems actually supported this module.  

The M9312 DEVICE flags should initially specify DEV_DIS as a default which makes it compatible with legacy configurations.
The logic in the m9312_reset routine:

1. should behave differently when the M9312 is disabled.  
2. should only allow the M9312 to be enabled for CPUs that supported it and otherwise it should actively set DEVICE_DIS.

The m9312_rd routine should always return NXM if the M9312 DEVICE is disabled.

In general, you should probably be using strcasecmp rather than strcmp,.  Additionally, rather than strncpy, we're migrating things to strlcpy.

Lastly, how is the ROM code in M9312 actually engaged by the user of the system?  How did original hardware end up running ROM code?

Is there really something called M93212?  You've used that name in several places (Initial Pull Request title and branch names in your Repo)...

- Mark

markpizz:
Jos,

After sleeping on this, I realized that Bob may think that the HITMISS changes may not be a good idea (I was looking only at the change set to the code rather than thinking through the value/implication of the change).  Bob may state that simh's goal is to be true to the hardware so that software that ran on the hardware still runs.  You may say that ROM based software is logically included since it ran on the hardware. 
 Well, maybe and maybe not.  Bob's goal of "running software" explicitly does not include running every diagnostic.  This is true since diagnostics very often test things timing with instruction loops or obtuse things about the original hardware which may have indeed been useful when the diagnostic was written but otherwise serve absolutely no purpose other than passing a particular diagnostic test and doesn't benefit any other software that ran on the system.  Due to this, numerous other ROM based diagnostics that run on simh simulators (the various VAX models in particular) have slightly modified ROM images which disable tests like this (cache behaviors in particular) and thus allow the other functions of the original ROMs to still be leveraged in simulation.  Given the 20 years the simh PDP11 has been running the many varied operating systems that actually ran on the original PDP11 hardware, implementing support for the cache diagnostic will likely be classified as something which should be patched in the ROM to be disabled...

Sorry I didn't think of this yesterday.  Meanwhile, looking through the M9312 listing details at https://www.ak6dn.com/PDP-11/M9312/, I don't see specific references to the HITMISS register which should be at 017745, right?  Where/how did you observe HITMISS failures?

- Mark

larsbrinkhoff:
Seeing HITMISS ran a faint bell for me.  I found #614, where a PDP-11/70 ROM requires the HITMISS register to be 1 for success.

larsbrinkhoff:
Isn't the existing ROM device (pdp11_rom.c) close or similar to the M9312?

markpizz:
> Seeing HITMISS ran a faint bell for me. I found #614, where a PDP-11/70 ROM requires the HITMISS register to be 1 for success.

How (under what conditions) did you observe this need?

> Isn't the existing ROM device (pdp11_rom.c) close or similar to the M9312?

I'm not sure yet, but part of what it is doing is specifically oriented around the many ROM images at  https://www.ak6dn.com/PDP-11/M9312/

larsbrinkhoff:
> How (under what conditions) did you observe this need?

My memory of this is very vague now, but I think it was some small self test before doing the main function of the ROM.

https://github.com/simh/simh/pull/614#issuecomment-418789508

markpizz:
> > How (under what conditions) did you observe this need?
>
> My memory of this is very vague now, but I think it was some small self test before doing the main function of the ROM.
> 
> #614 (comment)

So, the code that referenced the HITMISS register was part of the early ROM contents, and the solution which worked was merely to set the HITMISS register to 1, right?

larsbrinkhoff:
Yes, in this case just setting it to the static value 1 made the test pass.

markpizz:
Jos,

I see you're working on this stuff in your simh repo.  

Please participate in this discussion and answer the questions various questions that have been raised.

Thanks.

- Mark

friesga:
Mark, Lars,

Unfortunately my replies to your comments didn't reach this thread. The Github notification e-mails state that I can reply directly but that doesn't function. Sorry about the silence.

To my surprise a "dep hitmiss 1" indeed works and I therefore checked the execution of the diagnostic ROM code and the behaviour of simh. 

In TEST16 of the diagnostic code at address 165544 the following instructions are executed:
    ROR @177752  ; Check the HITMISS register if the last memory reference was a hit
    BCS 165556    ; Branch if yes

This instruction sequence is executed at least 1000 times for subsequent memory adresses and you would expect that the BCS would fail in the second pass of that code. In the subsequent passes however the HITMISS register still has the value 1. The reason for that is that the register is - as documented in the Processor Handbooks - read-only. The implementation in cpu70_wr() probably is correct: the the PSW C-bit is set whilst the HITMISS register is not modified and an SCPE_OK is returned. 

As the HITMISS register is read-only the simh "dep hitmiss 1" command however should fail I think.

I now see three options to make the diagnostic code pass the HITMISS register test:

1. Adapt the diagnostic code. Although this is an easy option, I personally wouldn't like it as I'm of the opinion that the simulator should be adapted to make original code function and not the other way around, adapt the code to make it run on a simulator.

2. Initialise the HITMISS register to 1 or make the cpu70_rd() function return 1, something like the implementation in CPUJ_rd(). Although not very nice, while properly documented I personally would find that an acceptable solution.

3. Implement the behaviour of the HITMISS register up to a point that it is accepted by the diagnostic code. Personally I would find that the cleanest solution and I now have - I think - a clean implementation ready.

Do you see other options? Which option do you prefer?

Regards, 

Jos

larsbrinkhoff:
(Note, I'm not replying in any SIMH official capacity.  Mark and Bob call the shots.)

I don't think the HITMISS register needs to be read-only as far as DEPOSIT goes.  The important thing is that executing PDP-11 code shouldn't be able to write it.

With regards to your points 1-2-3, recall that the stated goal of SIMH is first and foremost to run real operating systems and applications.  Diagnostics take a distinct second seat, and may not always be supported.  I think we should be happy that a simple `DEPOSIT HITMISS 1` makes this diagnostic run to completion.

friesga:
Hi Lars,

I'm sorry to say that - with all respect - I have to disagree with you. Using a "DEPOSIT HITMISS 1" to make the diagnostic code function is an option that ugly that I didn't even consider it. I fully agree with you (and Bob and probably Mark) that as Mark stated "'running software' explicitly does not include running every diagnostic". The 11/70 cache e.g. has two cache groups that can separately be enabled or disabled. The XXDP diagnostics enable and disable these groups thereby verifying that memory accesses accordingly produce hits or misses. Implementing that behaviour has no added value for simh and for me that's the point where an implementation of the HITMSS register should end. 

Being able to execute the diagnostic code that was built in the machines however is another story. In my opinion simh should reflect at the outside as much as useful and possible of the original behaviour. There are no patches made to e.g. drivers to make them run on simh and in my opinion m9312 diagnostic and boot ROMS have at least the same status. And this way "playing" with simh, making it as realistic as possible is fun, adapting original code to make it run om simh is a disappointment.

In that line I would also argue in favor of marking the HITMISS register read-only so a "DEPOSIT HITMISS" would return an error. Allowing a write on the registers causes confusion and put me on the wrong track.

Needless to say that I'll leave the verdict to Mark and Bob and will adapt my M9312 implementation to their choice.

markpizz:
I'm working with Bob on the HITMISS subject.

Meanwhile, I had asked:
> how is the ROM code in M9312 actually engaged by the user of the system? How did original hardware end up running ROM code?

Please answer this and provide potential configuration instructions to I can actually see the M9312 stuff in action.

friesga:
Mark,

My apologies for not answering all your questions, unfortunately my answers didn't reach the thread.

An actual (hardware) M9312 module is filled with ROMs in (two or more of) its five sockets. This makes the ROM code available at the (16-bit) addresses 176500, 173000, 173200, 173400 and 173600. In my m9312 implementation the sockets are filled with the simh command "SET M9312 SOCKET[0-4]=<ROM code>". ROM codes are e.g. B0 for the 11/60,70 Diagnostic and DL for a RL01/02 disk. The following commands e.g. configure a PDP-11/70 for booting from a RL02 disk:
simh> set cpu 11/70, 4M
simh> set m9312 socket0=b0
simh> set m9312 socket1=dp
simh> set rl0 rl02
simh> attach -r -e rl0 <disk image>

The ROM code is then booted just as in an original system by starting at 173004 (for a start without execution of the diagnostics) or at 173006 (for a start with execution of the diagnostics). The ROM code can also be started at 165744 with a code in the switch register to specify the device and unit to boot from. It all works exactly as documented in the "M9312 bootstrap/terminator module technical manual".

The added value of this boot procedure is that the system can now be booted the authentic way: by starting on an appropriate address in the ROM code. This is especially nice on Blinkenbone and PiDP systems as they can now be booted by a LOAD ADDRSS - START sequence.

larsbrinkhoff:
I for one think this effort is laudable.  My reason for adding the pdp11_rom device was to boot a GT40 in the "authentic" way with the VT11 ROM.  I'm wondering if your use case could be satisfied by adding one more slot to pdp11_rom (it only has four now).?  I see nothing so far in your description that says otherwise.

*If* (and only if) your M9312 device overlaps largely with my pdp11_rom, I think it would be wasteful duplication to have two ROM implementations.  I'm happy to accomodate changes to pdp11_rom if that's appropriate.

@friesga, could you send me, or point me to, the ROM image files?

There's the matter of HITMISS, but I think it's entirely orthogonal to the ROM device.

friesga:
Lars, 

My 9312 implementation indeed has some overlap with the ROM device. (Actually I used the ROM device to check that my approach was feasible). I agree with you that it would be better if there were one implementation for all ROM devices and I certainly would be willing to cooperate in implementing such a solution.

Our approaches however differ somewhat and it might be a challenge to merge their functionality. My approach for the m9312 implementation was to make its use as easy as possible for the users. I didn't want them having to search for ROM images and having these lying around somewhere and therefore included the ROM code as arrays built into the device (see pdp11_m9312.h in the pull request). I copied that approach from pdp11_xq_bootrom.h.

I we were to merge our two devices it should have the functionality of both devices, i.e. the M9312 ROMs built in plus the option to attach images as ROMs. Moreover, if our goal is to have just one ROM device, the design should be extensible to include possible future developments for other ROM devices too, like the M9301. Perhaps we could define a ROM layout parameter, indicating the number of sockets, the addresses at which the ROMS are accessible and the available ROMs per socket. "No layout" could then indicate the behaviour of the ROM device as it is. Would that be a feasible approach?

As source for the ROM code in the arrays I used the sources on Don North's site (https://www.ak6dn.com/PDP-11/M9312/). I assembled the sources and converted them to arrays by a modified version of Don's obj2bin (https://github.com/AK6DN/obj2bin). As a result of this approach there are no image files I can point you to.

The HITMISS matter is indeed more or less orthogonal to the m9312 implementation, but taking the effort for an implementation would be somehow useless if the ROM code then would fail to boot the system.Lars,

larsbrinkhoff:
Something like `set rom m9312` and `set rom m9301` etc should be feasible.  I imagine this would set the number, addresses, and memory contents of the sockets.  That is, if the part number uniquely identifies these details.

The maximum number of sockets could be raised to something that "should be enough for anyone", or dynamically allocated if possible.

I'll bow out of the HITMISS discussion.

markpizz:
> My 9312 implementation indeed has some overlap with the ROM device. (Actually I used the ROM device to check that my approach was feasible). I agree with you that it would be better if there were one implementation for all ROM devices and I certainly would be willing to cooperate in implementing such a solution.

We never really spoke about exactly what device @larsbrinkhoff's ROM device actually simulated and thus never explored the Qbus/Unibus constraints that may have existed with the actual hardware.  Maybe this was a M9301, right?  In any case, we need to revisit this and also add the additional constraint that either a M9301 OR a M9312 could exist in a system and not both, so it would seem most reasonable that a SET ROM M9312|M9301 would be appropriate.

> Our approaches however differ somewhat and it might be a challenge to merge their functionality. My approach for the m9312 implementation was to make its use as easy as possible for the users. I didn't want them having to search for ROM images and having these lying around somewhere and therefore included the ROM code as arrays built into the device (see pdp11_m9312.h in the pull request). I copied that approach from pdp11_xq_bootrom.h.

I believe that the built-in ROM images are absolutely most useful specifically for the reason @friesga states along with the fact that, these things are VERY stable and hardly ever modified (They're ROMs right!).  That said, there is also significant flexibility in allowing a developer (or user) to manually load a particular file and use it as a ROM's contents.  There are existing cases where this happens in current simulators: The various Qbus VAX simulators support explicitly loading a ROM image before booting the system with a fallback to loading the default (built-in) ROM image if a processor boot is attempted and the ROM hasn't yet been loaded.  Something like this should be supported here.  Meanwhile, since it is already known that an 11/60 & 11/70 want the B0 ROM, that should be the default when the CPU is one of these and the alternates should be set based on processor type and allow the user to override as desired.  It should certainly be easy enough to add a couple of GT40 and/or related ROM's to the set of built-ins which may then only be usable in M9301 mode...

> The ROM code is then booted just as in an original system by starting at 173004 (for a start without execution of the diagnostics) or at 173006 (for a start with execution of the diagnostics). The ROM code can also be started at 165744 with a code in the switch register to specify the device and unit to boot from. It all works exactly as documented in the "M9312 bootstrap/terminator module technical manual".

You mention the "switch register", which the "M9312 bootstrap/terminator module technical manual" indicates as a set of switches on the M9312 board.  It would seem that this should then be exposed as a simh REGister or at least settable by a SET <dev> SWITCH=xxx (without knowing more I'm thinking a simh REG is the best choice).  Meanwhile, once such a register exists, how does the code running from ROM access the contents of this switch register?

> The ROM code is then booted just as in an original system by starting at 173004 (for a start without execution of the diagnostics) or at 173006 (for a start with execution of the diagnostics). The ROM code can also be started at 165744 with a code in the switch register to specify the device and unit to boot from. It all works exactly as documented in the "M9312 bootstrap/terminator module technical manual".

I was only slightly a PDP11 user a few times and never had to boot one.  Are you saying that each time a PDP 11/70 booted someone had to flip a bunch of switches on the system front panel any number of times in order to achieve a boot?  There was no way to setup a system that would just boot to some specific device when you powered it on? Sort of like how the VAX 11/750 (or 730) had a knob to point to one of 4 boot device choices and a reset button on the panel which would cause a boot, and/or a switch on the front which would enable auto-boot from power on...

> The added value of this boot procedure is that the system can now be booted the authentic way: by starting on an appropriate address in the ROM code. This is especially nice on Blinkenbone and PiDP systems as they can now be booted by a LOAD ADDRSS - START sequence.

What is this "LOAD ADDRSS - START sequence."?

----
As for the HITMISS question, what you implemented didn't actually model the true hardware cache, but merely somehow produced different results from time to time.  Johnny Billquist test on real PDP 11/70 hardware showed a HITMISS register value of 000044.  This value clearly indicates that there are more bits of information/meaning in what it provides that reflect more complicated behaviors on the true hardware cache.  For simh uses.  it would seem sufficient to set the HITMISS register to 1 as a fudge value in the ROM/M9312 device's reset routine if the processor had actual cache (44/60/70).

markpizz:
@friesga, I see you asking on the simh list about the HITMISS for various systems with KDJ11-A and KDJ11-B boards.  Answers there may be interesting for the curious, but the answers aren't particularly relevant to a M9312 implementation.  

Bob Supnik reminded me of the Unibus specificity of the M9312.  Maybe there existed a Qbus ROM boards, which may be relevant to this effort, but not for running the code which specifically was designed for the Unibus systems.

friesga:
>> My 9312 implementation indeed has some overlap with the ROM device. (Actually I used the ROM device to check that my approach was feasible). I agree with you that it would be better if there were one implementation for all ROM devices and I certainly would be willing to cooperate in implementing such a solution.
 
> We never really spoke about exactly what device @larsbrinkhoff's ROM device actually simulated and thus never explored the Qbus/Unibus constraints that may have existed with the actual hardware. Maybe this was a M9301, right? In any case, we need to revisit this and also add the additional constraint that either a M9301 OR a M9312 could exist in a system and not both, so it would seem most reasonable that a SET ROM M9312|M9301 would be appropriate.

Agreed.
 
>> Our approaches however differ somewhat and it might be a challenge to merge their functionality. My approach for the m9312 implementation was to make its use as easy as possible for the users. I didn't want them having to search for ROM images and having these lying around somewhere and therefore included the ROM code as arrays built into the device (see pdp11_m9312.h in the pull request). I copied that approach from pdp11_xq_bootrom.h.
 
> I believe that the built-in ROM images are absolutely most useful specifically for the reason @friesga states along with the fact that, these things are VERY stable and hardly ever modified (They're ROMs right!). That said, there is also significant flexibility in allowing a developer (or user) to manually load a particular file and use it as a ROM's contents. There are existing cases where this happens in current simulators: The various Qbus VAX simulators support explicitly loading a ROM image before booting the system with a fallback to loading the default (built-in) ROM image if a processor boot is attempted and the ROM hasn't yet been loaded. Something like this should be supported here. Meanwhile, since it is already known that an 11/60 & 11/70 want the B0 ROM, that should be the default when the CPU is one of these and the alternates should be set based on processor type and allow the user to override as desired. It should certainly be easy enough to add a couple of GT40 and/or related ROM's to the set of built-ins which may then only be usable in M9301 mode...
 
That’s an interesting thought! I already was thinking about some kind of autoconfiguration for the ROMs. That might be challenging as there is just room for the boot code for four or five devices. (most ROMS support one device type, some support two) but I like challenges 😊.

>> The ROM code is then booted just as in an original system by starting at 173004 (for a start without execution of the diagnostics) or at 173006 (for a start with execution of the diagnostics). The ROM code can also be started at 165744 with a code in the switch register to specify the device and unit to boot from. It all works exactly as documented in the "M9312 bootstrap/terminator module technical manual".
 
> You mention the "switch register", which the "M9312 bootstrap/terminator module technical manual" indicates as a set of switches on the M9312 board. It would seem that this should then be exposed as a simh REGister or at least settable by a SET SWITCH=xxx (without knowing more I'm thinking a simh REG is the best choice). Meanwhile, once such a register exists, how does the code running from ROM access the contents of this switch register?
 
The switches determine the starting address in the ROM code and that determines the device to boot from. This could perform an auto boot a power up if desired.

>> The ROM code is then booted just as in an original system by starting at 173004 (for a start without execution of the diagnostics) or at 173006 (for a start with execution of the diagnostics). The ROM code can also be started at 165744 with a code in the switch register to specify the device and unit to boot from. It all works exactly as documented in the "M9312 bootstrap/terminator module technical manual".
 
> I was only slightly a PDP11 user a few times and never had to boot one. Are you saying that each time a PDP 11/70 booted someone had to flip a bunch of switches on the system front panel any number of times in order to achieve a boot? There was no way to setup a system that would just boot to some specific device when you powered it on? Sort of like how the VAX 11/750 (or 730) had a knob to point to one of 4 boot device choices and a reset button on the panel which would cause a boot, and/or a switch on the front which would enable auto-boot from power on...

The switches on the M9312 could be used to configure an auto boot on power up from a specific device, but that meant that the system had to boot always from the same device. In practice I never saw that option used and indeed someone “had to flip a bunch of switches on the system front panel any number of times in order to achieve a boot”.
 
>> The added value of this boot procedure is that the system can now be booted the authentic way: by starting on an appropriate address in the ROM code. This is especially nice on Blinkenbone and PiDP systems as they can now be booted by a LOAD ADDRSS - START sequence.
 
> What is this "LOAD ADDRSS - START sequence."?

That is setting the address to start at in the console switch register, pressing the LOAD ADRESS switch and then pressing the START switch. 

> As for the HITMISS question, what you implemented didn't actually model the true hardware cache, but merely somehow produced different results from time to time. Johnny Billquist test on real PDP 11/70 hardware showed a HITMISS register value of 000044. This value clearly indicates that there are more bits of information/meaning in what it provides that reflect more complicated behaviors on the true hardware cache. For simh uses. it would seem sufficient to set the HITMISS register to 1 as a fudge value in the ROM/M9312 device's reset routine if the processor had actual cache (44/60/70).

You’re completely right in your observation that my implementation doesn’t actually model a hardware cache, but it does produce a more realistic behaviour than just one fixed value. The value 044 in the HITMISS register indicates a HIT-MISS-MISS-HIT-MISS-MISS sequence in that order and my implementation could yield that result too. But I will be perfectly happen to have a fixed value if that will satisfy all ROM diagnostics.

markpizz:
> > You mention the "switch register", which the "M9312 bootstrap/terminator module technical manual" indicates as a set of switches on the M9312 board. It would seem that this should then be exposed as a simh REGister or at least settable by a SET SWITCH=xxx (without knowing more I'm thinking a simh REG is the best choice). Meanwhile, once such a register exists, how does the code running from ROM access the contents of this switch register?
> 
> The switches determine the starting address in the ROM code and that determines the device to boot from. This could perform an auto boot a power up if desired.

How does the PDP11 CPU know where to start execution on power-up in an auto-boot case?  

Related... Was an auto-boot model available on ALL or only some Unibus systems?

> > As for the HITMISS question, what you implemented didn't actually model the true hardware cache, but merely somehow produced different results from time to time. Johnny Billquist test on real PDP 11/70 hardware showed a HITMISS register value of 000044. This value clearly indicates that there are more bits of information/meaning in what it provides that reflect more complicated behaviors on the true hardware cache. For simh uses. it would seem sufficient to set the HITMISS register to 1 as a fudge value in the ROM/M9312 device's reset routine if the processor had actual cache (44/60/70).
> 
> You’re completely right in your observation that my implementation doesn’t actually model a hardware cache, but it does produce a more realistic behavior than just one fixed value. The value 044 in the HITMISS register indicates a HIT-MISS-MISS-HIT-MISS-MISS sequence in that order and my implementation could yield that result too. But I will be perfectly happen to have a fixed value if that will satisfy all ROM diagnostics.

How do you measure ROM diagnostic satisfaction?  Using the B0 and RL ROMS, hard setting a 11/70 HITMISS to 1 (by changing line 585 in pdp11_cpumod.c to:
``` 
*data = 1; //HITMISS;
```
and starting at 173006 lets the system successfully boot from the RL0 disk.  
Hard setting it to 0 causes a CPU HALT at 165554.  Are we expected to see some other sort of error message?

Using the MEM and RL ROMs, hard setting HITMISS to 1 or 0 on an 11/70 runs infinitely without producing output.  Is this expected?

As far as I can tell, the one fixed value seems to produce observably (to a normal user) equivalent results.  Maybe the ROM code is looping through different instructions but that is hardly interesting to a generic user running a simulator.

larsbrinkhoff:
@markpizz wrote:
> We never really spoke about exactly what device @larsbrinkhoff's ROM device actually simulated and thus never explored the Qbus/Unibus constraints that may have existed with the actual hardware. Maybe this was a M9301, right?

My primary concern was the ROM that comes with the VT11 in a GT40.  I don't know if the ROM is built into one of the VT11 boards, or if it's a separate board.  Noel Chiappa usually knows his PDP-11 stuff but his page here doesn't say: https://gunkies.org/wiki/VT11_Graphic_Display_Processor

I didn't have a single ROM type or bus in mind.  The ROMs just occupy a particular address range in the IO page, and the size depends on the file attached.

> In any case, we need to revisit this and also add the additional constraint that either a M9301 OR a M9312 could exist in a system and not both

I don't know, but with the proviso that knowledge of PDP-11 hardware is quite weak, I don't see any reason why multiple boards couldn't be installed.

larsbrinkhoff:
@friesga wrote:
> The switches determine the starting address in the ROM code

To clarify, are these switches on the ROM board (jumpers or DIP type switches)?  Hmm, yes I suppose they must be for auto boot.

(General information follows.)  This should not be confused with the front panel switches.  By toggling the switches to an address and pressing LOAD ADDRESS, and next START, you load an address into the PC and has the CPU go from there.

larsbrinkhoff:
If we are getting down to nitty gritty details or proposing updated ROM configurations, here is my current use case:

```
set rom enabled
set rom0 address=17766000
at rom0 out/pdp10-ka/bootvt.img
boot rom0
```

This places the ROM0 socket in the IO page and loads the code from a file.  The file size determines the address range.  `BOOT` is provided as a convenience but is much like DEPOSIT PC *ROM_start_address* followed by GO.

A suggested update would be to add `SET ROM TYPE=x` or `SET ROMn TYPE=x` as appropriate.  I think this should set the address, size, and contents.  The address could be overridden by using `SET ROMn ADDRESS`.

...

Now that I'm reading this http://gunkies.org/wiki/M9312_ROM I see a single M9312 type ROM board can have many different memory contents.  So just setting `TYPE=M9312` isn't enough.  It seems the M designation is the DEC module, or board?  And it could be fitted with many different ROM parts.  So maybe something like `SET TYPE=M9301-YA` or `SET TYPE=23-760A9` (see web page for part number), or `SET TYPE=RL02` to be more user friendly.

The M9312 has a jumper to set the address.  `SET JUMPER=x` could adress that, or `SET ADDRESS` explicitly.

The M9302 has a DIP switch, see https://gunkies.org/wiki/M9301_ROM.  `SET SWITCH=x`

larsbrinkhoff:
Here's detailed information about the M9312.  The photo clearly shows four ROM chips and a DIP switch.

https://www.pdp-11.nl/pdp11-34a/cpu/options/bootstrap-info.html

Here's a page about the M9302:

https://fritzm.github.io/m9301-troubles.html

friesga:
> @friesga, I see you asking on the simh list about the HITMISS for various systems with KDJ11-A and KDJ11-B boards. Answers there may be interesting for the curious, but the answers aren't particularly relevant to a M9312 implementation.

> Bob Supnik reminded me of the Unibus specificity of the M9312. Maybe there existed a Qbus ROM boards, which may be relevant to this effort, but not for running the code which specifically was designed for the Unibus systems.

Thanks for your comment Mark, I just want to be sure I understand all intricacies and uses of the HITMISS register. I don't want to contradict Bob, but maybe there is a connection with the M9312 as a11/84 (a UNIBUS-system) was equipped with both a KDJ11-B and M9312 compatible ROMs?

friesga:
>>> You mention the "switch register", which the "M9312 bootstrap/terminator module technical manual" indicates as a set of switches on the M9312 board. It would seem that this should then be exposed as a simh REGister or at least settable by a SET SWITCH=xxx (without knowing more I'm thinking a simh REG is the best choice). Meanwhile, once such a register exists, how does the code running from ROM access the contents of this switch register?
 
>> The switches determine the starting address in the ROM code and that determines the device to boot from. This could perform an auto boot a power up if desired.
 
> How does the PDP11 CPU know where to start execution on power-up in an auto-boot case?

That is determined by the switches on the M9312 board.

> Related... Was an auto-boot model available on ALL or only some Unibus systems?

As the auto-boot is a M9312 feature and the M9312 is a Unibus bootstrap/terminator module it would be available on all Unibus systems.

>>> As for the HITMISS question, what you implemented didn't actually model the true hardware cache, but merely somehow produced different results from time to time. Johnny Billquist test on real PDP 11/70 hardware showed a HITMISS register value of 000044. This value clearly indicates that there are more bits of information/meaning in what it provides that reflect more complicated behaviors on the true hardware cache. For simh uses. it would seem sufficient to set the HITMISS register to 1 as a fudge value in the ROM/M9312 device's reset routine if the processor had actual cache (44/60/70).
  
>> You’re completely right in your observation that my implementation doesn’t actually model a hardware cache, but it does produce a more realistic behavior than just one fixed value. The value 044 in the HITMISS register indicates a HIT-MISS-MISS-HIT-MISS-MISS sequence in that order and my implementation could yield that result too. But I will be perfectly happen to have a fixed value if that will satisfy all ROM diagnostics.
 
> How do you measure ROM diagnostic satisfaction? Using the B0 and RL ROMS, hard setting a 11/70 HITMISS to 1 (by changing line 585 in pdp11_cpumod.c to:
 
> ```
> *data = 1; //HITMISS;
> ```
> 
> and starting at 173006 lets the system successfully boot from the RL0 disk.
> Hard setting it to 0 causes a CPU HALT at 165554. Are we expected to see some other sort of error message?

That is the expected behaviour on a cache failure.
 
> Using the MEM and RL ROMs, hard setting HITMISS to 1 or 0 on an 11/70 runs infinitely without producing output. Is this expected?

I'll see if I can reproduce that and check the cause.
 
> As far as I can tell, the one fixed value seems to produce observably (to a normal user) equivalent results. Maybe the ROM code is looping through different instructions but that is hardly interesting to a generic user running a simulator.

I agree. As I pointed out in an earlier post that is thanks to the fact the HITMISS register is read-only and the ROR that is executed on the HITMISS register in the diagnostic code has no effect on the register itself.

friesga:
>> The switches determine the starting address in the ROM code
 
> To clarify, are these switches on the ROM board (jumpers or DIP type switches)? Hmm, yes I suppose they must be for auto boot.

You're right, for an auto-boot the address can be set in the switches on the M9312 module.

> (General information follows.) This should not be confused with the front panel switches. By toggling the switches to an address and pressing LOAD ADDRESS, and next START, you load an address into the PC and has the CPU go from there.

Completely right! That would be the "normal" use case.

markpizz:
> > >The switches determine the starting address in the ROM code
> >
> > To clarify, are these switches on the ROM board (jumpers or DIP type switches)? Hmm, yes I suppose they must be for auto boot.
> 
> You're right, for an auto-boot the address can be set in the switches on the M9312 module.

For auto booting on power up to work, setting the dip switches on the board to indicate the autoboot device is only part of the problem/solution.  Like I asked earlier, somehow the primary ROM code must be able to read the contents of these switches so that it can dispatch to the right device ROM with a specifier for the unit to actually boot.  It would seem that the contents of these switches somehow exist in the I/O page.  That detail needs to be understood and simulated.

> > How does the PDP11 CPU know where to start execution on power-up in an auto-boot case?
> 
> That is determined by the switches on the M9312 board.

Sounds like a chicken and egg answer here... :-)

There must be 2 things:
1) something physical configures the system to auto boot vs not.   What/how is that configured?
2) once auto boot is configured, something must somehow force the PC to point at the ROM and start execution there on power-up (equivalent to the manual LOAD ADDRESS ... START steps previously described).  How is this done?

larsbrinkhoff:
> By the time the J11 was retired, memory had become so fast, and so dense, that the CPU module could simply be populated with 4MB of fast RAM, and the "cache" never missed.

Did the ROM self test still work?  If the cache never misses, what does HITMISS contain?

markpizz:
I would guess that it contained 0177777 reflecting hits on every reference, but that is a SWAG.  This value, being non-zero does allow the M9312 boot diag to pass and invoke the device specific boot.

friesga:
> My primary concern was the ROM that comes with the VT11 in a GT40. I don't know if the ROM is built into one of the VT11 boards, or if it's a separate board. Noel Chiappa usually knows his PDP-11 stuff but his page here doesn't say: https://gunkies.org/wiki/VT11_Graphic_Display_Processor

@larsbrinkhoff  Is the source code of that ROM in machine readable form available somewhere? I found the GT40 Diagnostics Listing and I presume the code listed there is the bootprom code, but it's barely readable and a lot of work to type it over.

If it's available I could include it as a built-in ROM in my m9312/ROM device extension.

larsbrinkhoff:
I got the source code from here:  
http://www.brouhaha.com/~eric/retrocomputing/dec/gt40/software/bootrom/bootvt.s09

> I've recreated the GT40's bootstrap loader ROM source code from the listing in the maintenance manual. It doesn't match the ROMs in my GT40; I'm in the process of disassembling them and will present the results here when I'm finished.

This code is assembled as part of the [ITS](http://github.com/PDP-10/its) build.  The result is attached: 
[bootrom.zip](https://github.com/simh/simh/files/6124613/bootrom.zip)

friesga:
> I got the source code from here:
> http://www.brouhaha.com/~eric/retrocomputing/dec/gt40/software/bootrom/bootvt.s09

Great! Thanks Lars.

friesga:
Closed on issue of the renewed pull request.

PR #1014, opened 2021-02-27
AltairZ80: Fix showdata in MDS-AD device

psco:


markpizz:
Arrgh...  Somehow I missed this warning in Howard's recent commit to this file:

```
2>s100_mdsad.c
2>..\AltairZ80\s100_mdsad.c(507) : warning C4018: '<' : signed/unsigned mismatch
```

Since this PR is changing the same file so you want to fix it as part of this PR or another?

PR #1015, opened 2021-02-28
BESM6: Added emulation of magnetic tapes

leobru:
Added emulation of formatted BESM-6 tapes. Formatting is done automatically when a new file is attached to a device.
Formatting of blank tapes by the O/S does not work yet.

### DIFF
--- a/sim_video.c
+++ b/sim_video.c
@@ -448,6 +448,8 @@ if (vid_main_thread_handle == NULL) {
 
 vid_beep_setup (400, 660);
 
+memset (&event, 0, sizeof (event));
+
 while (1) {
     int status = SDL_WaitEvent (&event);
     if (status == 1) {


markpizz:
Hi Leo,

Several comments about the new besm6_mg:
1 - sum_with_right_carry and log_data are unused functions - no problem just FYI, 
2 - in several places, direct references to UNIT->fileref (specifically ferror(u->fileref)) should be avoided.  If there are structure or I/O problems with the internal tape format or problems with accessing the file data, the sim_tape_* API returning non success status should trigger error aborts.  The meaning of success should be contextually flexible (encountering a tape mark or end of media, etc. at appropriate times is success).
3 - in mg_attach, after success return from sim_tape_attach, you should use u->filename rather than cptr as the name of the attached file.  This is appropriate since, depending on flags set in sim_switches the cptr input sim_tape_attach may contain more than the simple file name that actually gets opened.
4 - in mg_attach, the "goto out;" could simply be "break;" and the out: label wouldn't be needed.  Since this logic is reporting an error, it would seem that rather than calling sim_printf to issue these error messages, you should call s = sim_messagef (SCPE_ARG,. "%s: filename must contain volume number 1..2047\n", sim_uname (u)) and then save the newly created tape name, and "sim_tape_detach(u);" and then "remove (saved_name);"  "return s;" rather than break.
5 - The subsequent "if (!sim_quiet && !(sim_switches & SWMASK ('Q')))" and then sim_printf could be reduced to: "sim_messagef (SCPE_OK, "%s: formatting tape volume %d\n", sim_uname (u), tapeno);"
6 - in mg_attach, the use of the SWMASK('N') bit in sim_switches is a little confusing.  It seems that you are trying to either open an existing tape image, or force newly created tapes to get "formatted".  These are reasonable goals, but this logic will allow essentially any already existing file to be attached regardless of whether it has an appropriate name OR if it has appropriate format data at the beginning of the tape.  Maybe a filename check belongs outside the while (1) loop, and the some logic which reads and validates the existing tape formatting information should exist, right?

\- Mark

leobru:
Hi Mark,

> 1 - sum_with_right_carry and log_data are unused functions - no problem just FYI,

I've modeled besm6_mg.c after besm6_disk.c, and forgot to cleanup; will do. sum_with_right_carry is specific to disk formatting.

> 2 - in several places, direct references to UNIT->fileref (specifically ferror(u->fileref)) should be avoided. If there are structure or I/O problems with the internal tape format or problems with accessing the file data, the sim_tape_* API returning non success status should trigger error aborts. The meaning of success should be contextually flexible (encountering a tape mark or end of media, etc. at appropriate times is success).

Noted; will do.

> 3 - in mg_attach, after success return from sim_tape_attach, you should use u->filename rather than cptr as the name of the attached file. This is appropriate since, depending on flags set in sim_switches the cptr input sim_tape_attach may contain more than the simple file name that actually gets opened.

Using u->filename may be fine, but for my purposes parsing the exact argument of the attach command will do just as well. I'm not treating it as a hierarchical file name anyway, I only need to fine the rightmost sequence of digits in it. And in the (theoretical?) case of versioning filesystems where the canonicalized file name might be appended with a suffix, like a version number in the form of "semicolon-number", or "@YYYY-MM-DD", or some such, parsing cptr as opposed to u->filename to extract the label number is a must.

> 4 - in mg_attach, the "goto out;" could simply be "break;" and the out: label wouldn't be needed. Since this logic is reporting an error, it would seem that rather than calling sim_printf to issue these error messages, you should call s = sim_messagef (SCPE_ARG,. "%s: filename must contain volume number 1..2047\n", sim_uname (u)) and then save the newly created tape name, and "sim_tape_detach(u);" and then "remove (saved_name);" "return s;" rather than break.
> 5 - The subsequent "if (!sim_quiet && !(sim_switches & SWMASK ('Q')))" and then sim_printf could be reduced to: "sim_messagef (SCPE_OK, "%s: formatting tape volume %d\n", sim_uname (u), tapeno);"
> 6 - in mg_attach, the use of the SWMASK('N') bit in sim_switches is a little confusing. It seems that you are trying to either open an existing tape image, or force newly created tapes to get "formatted". These are reasonable goals, but this logic will allow essentially any already existing file to be attached regardless of whether it has an appropriate name OR if it has appropriate format data at the beginning of the tape. Maybe a filename check belongs outside the while (1) loop, and the some logic which reads and validates the existing tape formatting information should exist, right?

What would you suggest should happen when a non-empty file not conforming to the SIMH tape format is attached? At the moment, it is treated as an unformatted tape: the OS will attempt to find a label and will fail, thus the unit will be inaccessible.

Allowing to attach existing files with arbitrary names is intentional. The numerical part of the file name is only needed when a new volume is being created. If switches could have arguments, the file name would not need to be parsed, and the attach command would have been either

1) attach [-E] device filename (here the file must exist with or without -E, otherwise the command will error out)
or
2) attach -N -L label device filename (discards the existing contents of the file, if any; formats the volume with the label)
or
3) attach -L label device filename (performs as (1) if the file exists, as (2) otherwise)

Lacking that mechanism, the numerical part of the file name is used.

markpizz:
> What would you suggest should happen when a non-empty file not conforming to the SIMH tape format is attached? At the moment, it is treated as an unformatted tape: the OS will attempt to find a label and will fail, thus the unit will be inaccessible.

If the OS is not actually capable of handling useful operations on "non conforming" tapes (i.e. those without the format initialized front-end), you should validate that format for pre-existing tapes at attach time so that the operator can immediately know about the error while interacting with SCP.  Otherwise., an arbitrary amount of time can exist between when the attach happens and when some other activity within the running system separately decides to attempt to access the tape.  This validation need not be related to the filename, but looking at the data in your formatting structure you should be able to find it or not.

> If switches could have arguments

Ahh...  Switches for tape attach DO have arguments (maybe not particularly useful for your situation), but this is precisely why to access the actual name of the file that has been opened you should parse u->filename which will then have all arguments extracted.

> the file name would not need to be parsed, and the attach command would have been either
> 
> attach [-E] device filename (here the file must exist with or without -E, otherwise the command will error out)
> or
> attach -N -L label device filename (discards the existing contents of the file, if any; formats the volume with the label)
> or
> attach -L label device filename (performs as (1) if the file exists, as (2) otherwise)
> Lacking that mechanism, the numerical part of the file name is used.

Well, there is a relatively large set of switches that are meaningful to sim_tape_attach (with more potentially being used in the future).  Currently these are: BCDEFLNVKQVX.  Lets say we reserve some switches for the sim_tape user, or you pick an unused switch name, or you decide that some particular currently used switch would never be useful to a user of your simulator.  Your code could:
1 - see the flag set in sim_switches, note that detail and remove it from sim_switches before calling sim_tape_attach and you would own its interpretation.  
2 - If interpreting this switch requires argument data. you can require its use to present arguments in your preferred order.  Note that  general simh argument parsing has all switches specified at the beginning of whatever cptr initially points at followed by each argument that may be implied by the respective switches.  The only place simulator authors encounter this is while parsing argument data in ATTACH commands, but the parsing paradigm is the same across all of SCP.  For example, say a command has switches A, B, and C that have no arguments, but switches D and E have require arguments with E (if specified) coming first.  These commands all follow these rules: and all present an identical sim_switches and cptr when SCP dispatches to the command:  
```
CMD -EDAC E-argument D-argument
CMD -ED -AC E-argument D-argument
CMD -E -D -A -C E-argument D-argument
CMD -ACDE E-argument D-argument
```
Lets say you pick Z as your flag to specify the label, the code would have:
```
t_bool label_specified = (sim_switches & SWMASK ('Z') != 0);
char label[CBUFSIZE];

sim_switches &= ~SWMASK ('Z');           /* remove label switch */
if (label_specified)
    cptr = get_glyph (cptr, label, 0);   /* get label */
if ((cptr == NULL) || (*cptr))
    return sim_messagef (SCPE_ARG, "Missing filename argument\n");
s = sim_tape_attach (u, cptr);           /* attach with remaining arguments */
```

PR #1026, opened 2021-03-28
makefile: Fix building with musl libc

Gorggg:
Linux systems with musl libc do not print anything upon invocation of ldconfig -p, making it so that the existing method of determining library search paths did not turn up paths that possibly contained SDL or libpcap libraries. This fixes such behaviour by automatically adding /lib and /usr/lib as library paths if they weren't found with ldconfig.

Thank you for looking at this. If there is anything I can clarify or improve, please let me know.

### DIFF
--- a/makefile
+++ b/makefile
@@ -371,6 +371,12 @@ ifeq (${WIN32},)  #*nix Environments (&& cygwin)
           endif
           ifeq (ldconfig,$(shell if ${TEST} -e /sbin/ldconfig; then echo ldconfig; fi))
             LIBPATH := $(sort $(foreach lib,$(shell /sbin/ldconfig -p | grep ' => /' | sed 's/^.* => //'),$(dir $(lib))))
+            ifeq (,$(filter /lib/,$(LIBPATH)))
+              LIBPATH += /lib/
+            endif
+            ifeq (,$(filter /usr/lib/,$(LIBPATH)))
+              LIBPATH += /usr/lib/
+            endif
           endif
         endif
         LIBEXT = so


markpizz:
1) You don't point at a system setup which actually does this, so I can't look at details.  Please elaborate on where this problem is actually observed.
2) Why is this a simh  problem and not a musl problem?  Linux link libraries are universally referenced specifically by the ldconfg paradigm.

Gorggg:
Thank you for your reply.

>     1. You don't point at a system setup which actually does this, so I can't look at details.  Please elaborate on where this problem is actually observed.

I first found this issue on Void Linux x86_64 musl. It is also present on Alpine Linux 3.13.3 x86_64.

>     2. Why is this a simh  problem and not a musl problem?  Linux link libraries are universally referenced specifically by the ldconfg paradigm.

In glibc-based systems, ldconfig is provided by glibc. Musl itself doesn't provide an ldconfig as described here: [https://wiki.musl-libc.org/faq.html](https://wiki.musl-libc.org/faq.html).
I'm sorry I can't speak to the full rationale behind this design decision. Various distributions provide a non-fully-featured ldconfig by symlinking to /bin/true or providing a script that can add library paths, but none of those that I have seen support the -p option. I wasn't able to find references to ldconfig in either the POSIX specifications or the Linux Standard Base, and the ldconfig implementations present on the BSDs don't support -p either. From what I can tell, ldconfig -p seems to be a glibc-specific feature, if you find that relevant.

markpizz:
Fixed in master branch with slightly modified solution.

PR #1035, opened 2021-04-19
video: unrecognized windowID

larsbrinkhoff:
Sometimes an SDL event is reported with windowID = 0 which is not a match to any window SIMH has opened.  Both @Isysxp and I have seen this.  I have no explanation for why this happens, but ignoring those events seem harmless.

I will produce a pull request that fixes this.

### DIFF
--- a/sim_video.c
+++ b/sim_video.c
@@ -380,11 +380,16 @@ static VID_DISPLAY vid_first;
 KEY_EVENT_QUEUE vid_key_events;                         /* keyboard events */
 MOUSE_EVENT_QUEUE vid_mouse_events;                     /* mouse events */
 
-static VID_DISPLAY *vid_window_from_id (Uint32 windowID)
+static VID_DISPLAY *vid_get_event_window (SDL_Event *ev, Uint32 windowID)
 {
 static Uint32 lastID = 0xffffffff;
 static VID_DISPLAY *last_display = NULL;
 VID_DISPLAY *vptr;
+SDL_KeyboardEvent *kev;
+SDL_MouseButtonEvent *bev;
+SDL_MouseMotionEvent *mev;
+SDL_WindowEvent *wev;
+SDL_UserEvent *uev;
 
 if (windowID == lastID)
     return last_display;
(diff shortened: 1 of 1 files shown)


markpizz:
2 things:
1) Please resubmit this PR based on the current head of the master branch.  You submitted 2 separate PRs both against the prior head of the master branch.  Avoiding a merge commit asks you to start again.
2) This change is fine, but I'm thinking that in the routine vid_window_from_id(), in the path that returns NULL, there should be some sim_debug logic that describes whatever is available about these events so that the potential exists in the future to understand their root cause.

larsbrinkhoff:
1. Rebased.
2. That's a good idea.  I'm not sure how to call sim_debug when there's no obvious device to point to.  I made something up, but maybe you have a better idea.

markpizz:
That debug approach doesn't work since there isn't any SCP mechanism to manage the debug state of an unconnected device and a device which only exists for this purpose really shouldn't exist.

Maybe the vid_window_from_id() should have its name changed to vid_get_event_window() and the argument should be a SDL_Event.  Merely emitting the window Id value in the debug output isn't particularly valuable.  I'm saying this so that that one place (where the ID isn't found) can display via sim_debug ALL of the event attributes for the event that wasn't found.  Otherwise the same logic would be needed by each caller that actually had the event structure available.  Meanwhile I'd suggest that the call to sim_debug should use vid_first->vid_dev as the DEVICE.  Maybe there should be a new SIM_VID_DBG_UNK_EVT debug symbol so that only this relatively rare (I hope) thing produces output which won't be entangled with other debug output unless you actually want it to be.

Isysxp:
This is quite a tricky problem.

  1.  The error seems to be Linux specific and I can only check SDL2 2.0.2 ...latest 2.0.12
  2.  The error is in SDL2 which reports an incorrect windowid in the event structure only for initial key events. Mouse events are correctly reported.
  3.  The windowid in the chained (vptr) list is correct.
  4.  There is no obvious alternative way of obtaining the correct windowid from the event structure as this does not contain a pointer to struct SDL_Window.
  5.  I suggest that SDL2 2.0.12 is checked to see if this error still occurs.

________________________________

markpizz:
I seem to remember seeing unexpected events when the thread digesting events got behind, and then there wasn't just one, but a failing case.

In any case, having the option to record unexpected events will be valuable if it happens again in the future.

> 2.  The error is in SDL2 which reports an incorrect windowid in the event structure only for initial key events. Mouse events are correctly reported.

If this is reliably reproducible with the initial event, then if it isn't fixed in the latest SDL2 version, then reporting the problem up stream should be done with a reproducible simple case (NOT simh based), digging that deeply may make an internal to SDL2 fix be obvious so providing a fix along with the report will pretty much guarantee that the fix makes it into the codebase.  I've done that with them in the past.

larsbrinkhoff:
Due to how SDL events are defined, the resulting code is a little unwieldy.  But here it is anyway.

`vid_first->vid_dev` may be NULL, so instead I used sim_messagef.

Isysxp:
For information:   The error has been corrected in SDL2 2.0.14 for Linux. Unpatched code is now functioning correctly.
                                 The version provided in windows-build (2.0.10) is also correct.
                                 It is unlikely that existing Linux distros will provide this version. A build/install cycle may be required.

________________________________

markpizz:
3 Things about this:

1) > vid_first->vid_dev may be NULL, so instead I used sim_messagef.

    I suggested debug since I got the impression that 1) this was a common occurrence and 2) it might produce excessive output.  If neither is true, then sim_messagef is OK.

2) If the event type is unknown (which is possible in this case of unexpected things), then you should at least emit the numeric value of the event type that was received so that it would be possible to track it down.

3) As @Isysxp points out:
    >    It is unlikely that existing Linux distros will provide this version. A build/install cycle may be required.
Since we know how a user might fix it, it would be useful for them to explore how to get the root problem fixed, so suggesting that SDL2 2.0.14 is needed absolves simh of the problem.  Users can then either pressure their supplier to update the packaged version or somehow deal with upgrading themselves.  Building with self built & installed packages is not supported by the makefile, but we'll cross that bridge when it gets there.

larsbrinkhoff:
I made some updates as per the suggestions.

I have tested SIMH, but I haven't seen the problem.  @Isysxp, do you have a way to reproduce it and see if my added code is triggered?

Isysxp:
Re: commit ab48f31<https://eur03.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fsimh%2Fsimh%2Fcommit%2Fab48f31317f63943cf36995d2e29feb16d122b36&data=04%7C01%7Cian.schofield%40newcastle.ac.uk%7Cef78a6dca0ee498dab4e08d9041ab701%7C9c5012c9b61644c2a91766814fbe3e87%7C1%7C0%7C637545335054910632%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=cvDA6w4twtmBMhR5eIGRnqQSsZ%2BGvhVtZ9Zxumb8FnY%3D&reserved=0> video: Ignore events from unrecognized windows.

I can confirm that this commit detects the error. It does not appear to correct the error.
Linux (Debian 8) SDL2 2.0.2
NB: Deprecated workaround. Minimise and restore graphic window.

________________________________

markpizz:
The goal was not really to correct the error, just to potentially ignore/report the internal SDL problem.  You said that it happens on the first keyboard event.  It would seem that a single report is reasonable in this case.  The original suggestion to only emit messages via debug output was an idea to he;lp development understand the issue while also ignoring such events.  If things don't work correctly after ignoring this event, but can then be fixed by minimizing and restoring the graphics window, then that suggestion should be part of what is emitted as well as the SDL2 version upgrade information.

markpizz:
FYI: Ubuntu 21.04 includes libsdl2-dev 2.0.14

PR #1044, opened 2021-05-09
3B2: Rename files in prep for Rev 3 support

sethm:
This change lays the groundwork for adding support for Rev 3 3B2 models,
which includes the Model 500, Model 600, and Model 1000.

Rather than use the fixed strings "400" and "1000" in file names, the
strings "rev2" and "rev3" will be used, which allows greater flexibility
to implement various system configurations more easily.

Additionally, this change adds a copy of the Debug Monitor (DEMON) ROM
image for the Rev 2 board, to be used soon in a later checkin.


markpizz:
Please change the line endings on the 3 new rom_rev2_xxx_bin.h files to CRLF to be consistent with the rest of the source code line endings...

markpizz:
Actually, it currently emits "text" files (since the call to fopen merely says "w").  That produces LF line endings on Unix systems and CRLF line endings on Windows.  

To avoid this potential in the future, I'll change the fopen mode to "wb" and change the handful of nearby places which previously had \n to have \r\n instead.

Meanwhile, can you squash these two commits into one and push -f the branch to your repo so there's one commit for this change.

PR #1055, opened 2021-07-26
wd179x: Add support for wd1795/7 side select output.

hharte:
@psco, @markpizz : FYI, this adds support for the WD1795 / 1797 side select output.


psco:
This looks fine. Thank you Howard.

PR #1066, opened 2021-08-09
3B2: Rev 3 Development Base

sethm:
This change introduces initial support for the AT&T 3B2 Rev 3 platform, based
around the WE32200 CPU with up to 64MB of RAM and SCSI disk and tape support.

This simulator is experimental and not yet supported. It will not be built by
default, but can be built with:

     make 3b2-600

Or by using the 3B2-600 Windows Visual Studio project.


psco:
Xcode on MacOS chokes on the angled brackets
`#include <sim_defs.h>` (line 46 in 3b2_cpu.c)
and
`#include <scp.h>` (line 33 in 3b2_rev3_mmu.c)

Changing this to
`#include "sim_defs.h"`
resp.
`#include "scp.h"`
solves the problem.

markpizz:
Actually, less is better.

1) Remove "#include <sim_defs.h>" from 3b2_cpu.c, remove "#include <scp.h>" from 3b2_rev3_mmu.c

This should solve the XCode problem.  @pcsco, it would be nice if you could show me how to setup the XCode case that failed on MacOS on my environment.  Simple compiling with make from a terminal window works without error.  I've got XCode installed so I should be ready to go.

2) '#include "sim_defs.h"' really shouldn't be done anywhere except one form or another of 3b2_defs.h (or the model specific variants of that 3b2_rev2_defs.h).   Any current specific cases of "#include "sim_defs.h" in some include files should probably be replaced by '#include "3b2_defs.h"'

sethm:
Yes, agreed, I will pull the two offending includes out, as well as removing the unnecessary sim_defs.h includes. They don't belong there and they serve no purposes. I'll prepare a PR.

PR #1069, opened 2021-08-11
3B2: Header cleanup

sethm:
Remove unnecessary include directives and improve include hygiene.

### DIFF
--- a/3B2/3b2_cpu.c
+++ b/3B2/3b2_cpu.c
@@ -42,15 +42,27 @@
  * - "WE 32200 Microprocessor Information Manual" (AT&T, 1988)
  *
  */
-#include <sim_defs.h>
 
-#include "3b2_defs.h"
 #include "3b2_cpu.h"
+
 #if defined(REV3)
+#include "3b2_rev2_mau.h"  /* TODO: Replace with Rev 3 MAU when implemented */
+#include "3b2_rev3_csr.h"
+#include "3b2_rev3_mmu.h"
 #include "rom_rev3_bin.h"
 #else
+#include "3b2_rev2_csr.h"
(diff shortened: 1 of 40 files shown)


markpizz:
I'm thinking you may have gone too far with removing includes.  For example, in 3b2_cpu.c you removed 3b2_cpu.h (and similarly for numerous other .c files).  I've found that having the module's include file included in the source it defines interfaces for provides the easiest place for declaration inconsistencies to be noticed and flagged by compilers.  Without this, on some platforms link time checking might catch such inconsistencies, while on others, nothing may be noticed and you're left with runtime issues which may send you into the weeds...

sethm:
I agree it would be worth taking another pass at this, but it's a larger change than I'll be able to make in the next day or two. Unfortunately, the 3B2 headers are structured poorly right now and I've found it's very tricky not to introduce cycles. This can be fixed, of course, but it will involve some thought so it's done correctly.

I'll keep this PR open and work on the changes, making sure that my master branch in fully in sync with upstream master before I force-push a new update.

markpizz:
well, I just looked closer at what you're doing and as long as you include ALL of the modules' .h files in 3b2_defs.h, then the interfaces for each .c file will absolutely explicitly be declared and thus visible at compile time so that inconsistent declarations vs implementations can be reported.  I'm going to merge this now so that the MacOS XCode stuff gets in.  If you want to reorganize the includes later go ahead and submit a pull request at that time.

sethm:
Damn, it looks like I raced you, I'm sorry about that.

If you prefer, I'll back out the change I just pushed and JUST fix the XCode compilation issue, with no other changes.

PR #1071, opened 2021-08-23
3B2: Interrupt subsystem refactor

sethm:
1. This change introduces a full refactor of the interrupt subsystem
   for the system board (SBD) and the I/O bus (CIO).  Interrupt decode
   should now be significantly faster, and not require an expensive
   calculation on every step.

2. The TIMER device has been split into Rev 2 and Rev 3
   implementations.

3. The DEMON Rev 2 Debug Monitor ROMs can now be optionally compiled
   into the 3B2/400 simulator by passing "-DDEMON" at compile time.
   Note that this will require also adding "TESTS=0" to avoid trying
   to run the compile-time 3B2/400 diagnostic test harness, which is not
   expecting the DEMON ROM.


markpizz:
Fix this First:

1>3b2_rev3_timer.c
1>..\3B2\3b2_rev3_timer.c(125) : warning C4101: 't' : unreferenced local variable
1>..\3B2\3b2_rev3_timer.c(218) : warning C4101: 'time' : unreferenced local variable
1>..\3B2\3b2_rev3_timer.c(269) : warning C4101: 'time' : unreferenced local variable

sethm:
The unreferenced local variable warnings are now fixed following my force-push.

markpizz:
Thanks for the compile cleanup.

Meanwhile, I just paid enough attention to your commit comments for this.  Specifically:

> The DEMON Rev 2 Debug Monitor ROMs can now be optionally compiled
> into the 3B2/400 simulator by passing "-DDEMON" at compile time.
> Note that this will require also adding "TESTS=0" to avoid trying
> to run the compile-time 3B2/400 diagnostic test harness, which is not
> expecting the DEMON ROM.

For all simulators I've made specific effort to avoid compile time inputs that change behavior or functionality.  The goal would be for no user to have to recompile the simulators with different options to get different functionality.  Such options should usually be selected during simulator runtime by some sort of SET command or possibly different arguments to a BOOT command.

As far as implementing such an option (via SET or BOOT) it seems slightly complicated when using BuildROMs created include files, but not particularly hard.  The VAX750 simulator has two dynamically alternate boot ROMs and they're both included in vax750_mem.c.  It includes two boot rom include files using "#undef BOOT_CODE_SIZE", etc. between the respective includes.  The BOOT command for this simulator parses it's argument and based on the value uses one of the ROMs which are distinguished by the array name that each respective rom is known by.  

However, to make multiple alternate ROM support somewhat easier, I've just reworked BuildROMs to define unique #define symbols (BOOT_CODE_SIZE_n, BOOT_CODE_FILENAME_n, BOOT_CODE_ARRAY_n) in the case where multiple ROM includes are performed.  No need to #undef anything.  These #defined symbols could then be used to potentially populate an array initialization.

markpizz:
Arggh...  

I broke the 3b2 build with the change to the BuildROMs output.  It was broken due to the fact that in 3b2_rev2_mmu.h you had defined the BOOT_CODE_SIZE name and the changed ROM definition files now explictly only define BOOT_CODE_SIZE, BOOT_CODE_FILENAME and BOOT_CODE_ARRAY if BOOT_CODE_SIZE isn't defined.  This check is needed to avoid defining the same names with different values and causing a compile error.

I didn't catch this when I tested since I was working in a directory which had the commit for this PR and you must have stopped doing defining BOOT_CODE_SIZE anywhere.

I'll change the BuildROMs generated files to avoid the above mentioned check and just explicitly #undef each name before defining it while continuing to generate the BOOT_CODE_SIZE_n and related names.

sethm:
I'll get my master branch cleaned up and in sync with upstream master, and then I'll redo the multiple ROM support.

markpizz:
The build problem is now fixed.

Things will be stable for at least a day from my side.

Do you want to drop this PR and come back with a new one?  It doesn't matter to me, but the opening comment in this PR will no longer be valid...

sethm:
Yup, I'll kill this PR and come back with a new one. Thanks.

PR #1072, opened 2021-08-24
3B2: Interrupt Refactor and DEMON ROM

sethm:
- This change introduces a full refactor of the interrupt subsystem
  for the system board (SBD) and the I/O bus (CIO). Interrupt decode
  should now be significantly faster, and not require an expensive
  calculation on every step.

- The TIMER device has been split into Rev 2 and Rev 3
  implementations.

- The optional 3B2/400 Debug Monitor ROMs can now be booted by passing
  the "DEMON" argument to the 3B2/400 simulator BOOT command. Any
  of the following will cause the Debug Monitor ROM to be booted
  instead of the standard 3B2/400 ROM:

       sim> BOOT DEMON
       sim> BOOT CPU DEMON
       sim> BOOT DEMON CPU


markpizz:
What are your thoughts about this warning:

```
In file included from ./3B2/3b2_rev3_mmu.c:31:
./3B2/3b2_rev3_mmu.c: In function ‘void put_sdce(uint32, uint32, uint32)’:
./3B2/3b2_rev3_mmu.h:170:37: warning: suggest parentheses around arithmetic in operand of ‘|’ [-Wparentheses]
  170 | #define SD_TO_SDCH(hi,lo)     ((hi) & SD_ADDR_MASK           |  \
      |                                     ^
./3B2/3b2_rev3_mmu.c:187:26: note: in expansion of macro ‘SD_TO_SDCH’
  187 |     mmu_state.sdch[ci] = SD_TO_SDCH(sd_hi, sd_lo);
      |                          ^~~~~~~~~~
```
If one set of parentheses are appropriate, then many would be for all expressions in 3b2_rev3_mmu.h from lines 171 thru 211.  It is up to you.

sethm:
I think it's a valid warning. I was only able to get it by compiling with GCC-11, oddly enough, but it's a good catch. I've added the parentheses and pushed, so I'm happy with the state of the macros in that file now.

sethm:
Hmm, the vax8600 build seems to have failed diagnostics. Nothing I've touched should have interacted with other simulators, and I'm completely unfamiliar with the VAX diagnostics, so I'll let you look into it.

PR #1074, opened 2021-08-29
VIDEO: Fix bug in vid_dst_last handling.

larsbrinkhoff:
The handling of vid_dst_last and vid_data_last was not updated to work with multiple windows, which can cause window updates to be lost.

I'm posting this draft which fixes the problem by adding a new vid_vptr_last to keep track of the latest window making an update.  I'm not sure if this is preferable, or of each window should maintain their own vid_dst and data_last information.

### DIFF
--- a/sim_video.c
+++ b/sim_video.c
@@ -371,6 +371,8 @@ DEVICE *vid_dev;
 t_bool vid_key_state[SDL_NUM_SCANCODES];
 VID_DISPLAY *next;
 t_bool vid_blending;
+SDL_Rect *vid_dst_last;
+uint32 *vid_data_last;
 };
 
 SDL_Thread *vid_thread_handle = NULL;                   /* event thread handle */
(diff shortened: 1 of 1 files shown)


markpizz:
Why wouldn't each window have their own 'last' objects?  What's going on in one window is unrelated to what's going on in another, right?  Additionally the data in the 'last' objects are protected by a per window mutex.  Since there are different mutexes, there should be different 'last' objects.

larsbrinkhoff:
Sounds good.  I'll update the pull request.

markpizz:
On a related note.  I see that the current code is allocating and freeing these 'last' entries for each draw operation that happens.  Since draw operations happen very often the overhead of malloc and free for these things can be avoided by allocating them once and keeping the filled state in some sort of flag rather than merely the fact that the pointer is not NULL.  The buffer data is variable in size, so a separate variable would keep track of the allocated size and if, when it is to be filled, it is too small, it can be realloced and the buffer size updated.

larsbrinkhoff:
I'm on it.

larsbrinkhoff:
I updated to make the "last" data per-window.

I can take a look at the allocation optimization next.

markpizz:
In vid_draw_window(), the variable last is assigned a value WITHOUT being protected by the vid_draw_mutex,  That assignment should be done after the Mutex is locked.

PR #1085, opened 2021-10-18
H316: Fix host interface word counts.

larsbrinkhoff:
The host interface uses the UDP packet format from the modem interface, with an additional flags word.  Ensure the flags are not included in the word count.  @charlesUnixPro, please check my reasoning is correct.

### DIFF
--- a/H316/h316_hi.c
+++ b/H316/h316_hi.c
@@ -344,7 +344,7 @@ void hi_start_tx (uint16 line, uint16 flags)
     tmp [0] = flags;
     for (i = 0; i < count; i ++)
       tmp [i + 1] = M [next+i];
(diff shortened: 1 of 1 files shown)


charlesUnixPro:
On Mon, Oct 18, 2021 at 5:28 AM Lars Brinkhoff ***@***.***>
wrote:

> The host interface uses the UDP packet format from the modem interface,
> with an additional flags word. Ensure the flags are not included in the
> word count. @charlesUnixPro <https://github.com/charlesUnixPro>, please
> check my reasoning is correct.
>
Modulus that it has been a long time since I looked it the code, the fix
looks correct to me.

-- Charles

PR #1090, opened 2021-10-30
"branchless" condition code evaluation

bscottm:
[You might wonder if "the juice is worth the squeeze." Generally speaking, yes, if conditional branches are unpredictable (50/50 odds of taking either path).]

High performance computing hack: Use boolean test results (0/1) to compute the condition code results, as opposed to if/else constructs (part of STD99.) For x86, there are fewer (sometimes no) conditional jumps, which reduces branch prediction delays and increases the length of basic blocks (instruction sequences between jumps).  Conditional branches are replaced with code sequences that utilize 'cmov*' instructions. Extract from the CMPB code:

````
CC_BRANCHLESS (gcc)              |!CC_BRANCHLESS (gcc)
.L1741:                          |.L1741:
        movl    -68(%rbp), %r8d  |        movl    -68(%rbp), %edx
        movl    -72(%rbp), %ecx  |        movl    -72(%rbp), %eax
        movl    $8, %eax         |        cmpb    %al, %dl
        cmpb    %r8b, %cl        |        jg      .L2702
        cmovge  %edi, %eax       |        je      .L3028
        cmpb    %cl, %r8b        |.L1960:
        movl    %eax, %edx       |        cmpb    %al, %dl
        movl    $4, %eax         |        jbe     .L1961
        cmove   %eax, %edi       |        orl     $1, %r15d
        seta    %al              |.L1961:
        movzbl  %al, %eax        |        movl    %r15d, -232(%rbp)
        orl     %edx, %eax       |        jmp     .L1348
        orl     %eax, %edi       |.L3028:
        movl    %edi, -232(%rbp) |        movl    $4, %r15d
        jmp     .L1348           |        jmp     .L1961
                                 |.L2702:
                                 |        movl    $8, %r15d
                                 |        jmp     .L1960
````

This hack also works for setting the overflow trap and computing flags for the VAX CMP* instructions.

Caveat: While the code eliminates extra jumps, the native pipeline can serialize waiting for condition codes from the previous instruction. Serializing the native pipeline is generally preferable to stalling in branch prediction.

Static inline code is used to minimize the preprocessor conditional compilation (ref: `vax_mmu.h`, which also has static inlines for similar purposes.)

vax4xx_va, vax_va.c: remove duplicate extern declarations for trpirq.
vax_cpu1.c: Fix signed/unsigned mismatch for trpirq.

### DIFF
--- a/VAX/vax4xx_va.c
+++ b/VAX/vax4xx_va.c
@@ -127,7 +127,6 @@
 extern int32 int_req[IPL_HLVL];
 extern int32 tmxr_poll;                                 /* calibrated delay */
 extern int32 fault_PC;
-extern int32 trpirq;
 
 uint16 va_ram[RAM_SIZE];
 
--- a/VAX/vax_cpu.c
+++ b/VAX/vax_cpu.c
@@ -237,7 +237,7 @@
 uint32 *M = NULL;                                       /* memory */
 int32 R[16];                                            /* registers */
 int32 STK[5];                                           /* stack pointers */
-int32 PSL;                                              /* PSL */
+uint32 PSL;                                             /* PSL */
 int32 SCBB = 0;                                         /* SCB base */
 int32 PCBB = 0;                                         /* PCB base */
 int32 P0BR = 0;                                         /* P0 mem mgt */
@@ -250,7 +250,7 @@ int32 SISR;                                             /* swre int req */
 int32 ASTLVL;                                           /* AST level */
 int32 mapen;                                            /* map enable */
 int32 pme;                                              /* perf mon enable */
-int32 trpirq;                                           /* trap/intr req */
+uint32 trpirq;                                          /* trap/intr req */
 int32 in_ie = 0;                                        /* in exc, int */
 int32 recq[6];                                          /* recovery queue */
 int32 recqptr;                                          /* recq pointer */
@@ -491,7 +491,8 @@ return buf;
 
 t_stat sim_instr (void)
 {
-volatile int32 opc = 0, cc;                             /* used by setjmp */
+volatile int32 opc = 0;                                 /* used by setjmp */
+volatile uint32 cc;                                     /* used by setjmp */
 volatile int32 acc;                                     /* set by setjmp */
 int abortval;
 t_stat ret;
@@ -1897,8 +1898,12 @@ for ( ;; ) {
         r = (op1 + op0 + (cc & CC_C)) & LMASK;          /* calc result */
         WRITE_L (r);                                    /* store result */
         CC_ADD_L (r, op0, op1);                         /* set cc's */
+#if defined(CC_BRANCHLESS)
+        cc |= ((r == op1) && op0) * CC_C;
+#else
         if ((r == op1) && op0)                          /* special case */
             cc = cc | CC_C;
+#endif
         break;
 
     case ADDL2: case ADDL3:
@@ -1923,8 +1928,12 @@ for ( ;; ) {
         r = (op1 - op0 - (cc & CC_C)) & LMASK;          /* calc result */
         WRITE_L (r);                                    /* store result */
         CC_SUB_L (r, op0, op1);                         /* set cc's */
+#if defined(CC_BRANCHLESS)
+        cc |= ((op0 == op1) && r) * CC_C;
+#else
         if ((op0 == op1) && r)                          /* special case */
             cc = cc | CC_C;
+#endif
         break;
 
     case SUBL2: case SUBL3:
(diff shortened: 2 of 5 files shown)


markpizz:
This change has similar consequences/complexity to the recent simh discussion and previous efforts to make the vax simulator uint32 clean and avoiding the consequences of some of the C99 behaviors that compilers now do.

The scope of the change is large enough to require rigorous testing to prove that the resulting VAX is still a VAX.  Being able to run the basic VAX 780 / 8600 diagnostics is certainly a necessary condition to incorporate this or the previously mentioned changes, however those diagnostics are do not sufficiently exercise the instructions with varying and complex data that DEC did when it verified each new processor model.  The tool they used to achieve this is/was known as AXE.   If you can inspire someone at HP to either provide the AXE tool so we can run it ourselves or provide a means where someone at HP can run it for us and provide feedback about what works and what doesn't, then we can embrace this change and varying pieces of other work that tries to bring the VAX simulator up to the subtleties in the generated results of new compilers.

bscottm:
0/1 boolean values have been part of STD99 through the present, so it's not a STD99-specific feature. That's why it passes the tests on the Github buildbots.

uint32-clean is a big change; I've been playing with that for over a year -- the vax_cpu/cpu1 changes aren't big, but the surrounding infrastructure in PDP-11 and the underlying simh core are fairly extensive.

One way around the AXE problem for this pull request would take the form of using the equivalent of Haskell's [`QuickCheck`](https://hackage.haskell.org/package/QuickCheck). Functionally, the two condition code evaluations are the same, Could be fairly straightforward to prove that with a bit of scaffolding. That's something I could put together on the CMake branch, since CMake's `ctest` structure makes that easy.

archivisth:
This is not worth the risk. The performance gain will be minor, the potential consequences in the form of undiscovered bugs major.

There are individuals and organizations that depend on SimH VAX, in its validated form, to run their old but still critical systems.

This should be put on a side branch and not on the main VAX branch.

bscottm:
@archivisth I'd be happy if it were on a branch and let people "kick the tires" to gain confidence. Had I written functionally inequivalent code, the issue gets picked up before EVKAA executes, during the basic core instruction testing (something I actually tried.) Granted the test suite is necessary, but insufficient, it does provide some level of confidence that either code choice is correct at a functional level.

markpizz:
> I'd be happy if it were on a branch and let people "kick the tires" to gain confidence.

I'm 100% sure that when new VAX processors were produced by Digital, the development engineering folks were 100% "confident" that what they'd produced was indeed a VAX and it ran normal operating system activities and the basic diagnostics successfully.  The company then used AXE to actually confirm the fact or identify potential problems.  Bob may have some insight relating to how many confident iterations actually occurred for various projects he was involved in before they actually shipped.

archivisth:
In 1991, I compiled a detailed analysis of bugs in the four VAX microprocessor designs - MicroVAX II, CVAX, Rigel, NVAX.

Total passes+minipasses (CVAX includes CMOS-2 shrink bug fix pass):

	uVAX		CVAX		Rigel		NVAX
	----		----		-----		----
	5+11		5+11		6+1		2+0


Bugs found per pass:

	pass:	1	2	3	4	5
		-	-	-	-	-
uVAX		 6	11	10	3	3
CVAX		19	 7	 7	3
Rigel		 7       4	 1	2	2
NVAX		15	 3


Breakdown of bugs by error type:

	type:	logic	circuit	layout	ucode	spec change
		-----	-------	------	-----	-----------
uVAX errors	  9	   17	   2	  5	    0
uVAX fixes	  7	   16	   2	  5	    2

CVAX errors	 15	    9	   5	  7	    0
CVAX fixes	 15	    9	   5	  6	    1

Rigel errors	  3	    4	   0	  9	    0
Rigel fixes	  3	    4	   0	  8	    1

NVAX errors	 11	    1	   1	  5	    0
NVAX fixes	  8	    1	   1	  8	    0

In my conclusion, I wrote "The significantly increased verification efforts on Rigel and NVAX were repaid by significantly reduced bug rates." Key to that was being able to run <some> AXE cases on Rigel during simulation (on CVAX servers and workstations), and a <greater> number of AXE cases on NVAX (on Rigel servers and workstations). Nonetheless, a VAX was not considered shippable until it had run a minimum of 5 million cases in each instruction class (roughly 60 million total for a subset VAX). Until SimH, which is as fast as or faster than most VAXes, this required real hardware. Note that AXE production runs required the hardware to be running VMS - that is, running VMS was not, in and of itself, sufficient verification for a VAX.

markpizz:
I just updated @archivisth message to format the tables so the columns are properly aligned.

PR #1099, opened 2021-12-10
Initial support for Meson buildsystem

XVilka:
To simplify compilation and cross-platform support, and inspired by various C projects that switched from autotools to Meson, including QEMU, I did an initial work of migrating from the Makefile to Meson.

There are two implementations of this language:
- [Meson](https://mesonbuild.com)
- [Muon](https://sr.ht/~lattis/muon/) - mostly work-in-progress

For now, I did the simplest compilation without running tests and optional features. Please let me know if there is an interest in this, before I proceed.

I split in two commits - adding the buildsystem and removing the unnecessary files.
The second commit is not really necessary, since these can coexist without any conflict, making the gradual transition possible, like X.org, Wayland, Mesa, QEMU did, slowly deprecating autotools on the course of multiple releases.


markpizz:
Just because something doesn't happen to be how you may have solved the problem, doesn't mean there is a glaring need to make changes merely for change sake.  "If it ain't broke, don't fix it!".

Feel free to keep your own methods to things your way, but those changes will live in your personal repo.

Meanwhile, if you have contributions which add useful/missing functionality to existing simulators or better yet write a new simulator we will be glad to include those efforts.

PR #1103, opened 2021-12-25
TMXR, PDP11, PDP10, VAX: Add DDCMP sync framer support

pkoning2:
This adds support for the "framer" device, which is a USB-connected
device built around a Raspberry Pico that connects to a synchronous
line, either RS-232 or DEC "integral modem" coax connection.  It
implements the framing portion of DDCMP: clock recovery for the
integral modem case, byte sync, and DDCMP frame handling including
CRC.  The actual DDCMP protocol state machine, with its handling of
sequencing, timeout and retransmit, etc. is left to the host
software.  All the design files for the framer may be found at
https://github.com/pkoning2/ddcmp .

This commit adds code to drive the framer from the TMXR library,
allowing it to be used either from simulated DMC-11 or simulated
DUP-11 devices.  Both have been tested, using RSTS/E, RSX-11/M+, and
TOPS-20.


markpizz:
I'm looking at this now.

1) Cosmetic, but required issue relates to consistency in indentation style.
2) Parsing of attach parameters should reject incompatible options.
3) Framer speed should leverage existing speed validation and computation logic (extended as necessary for different framer requirements).
4) Device simulation code (not help comments) should know nothing about framer.  tmxr_start_framer and tmxr_stop_framer should be internal to TMXR.  You'll say: but, tmxr_start_framer  allows for specifying some sort of DMP specific mode.  Elaborate on this detail.  BTW, although DMP device simulation belongs in pdp11_dmc.c, currently only some of the basic things are actually implemented with no actual functioning device.  tmxr_start_famer and tmxr_stop_framer could happen at attach time (in TMXR), or on DTR transitions.

I've reproduced the problem Reindert was seeing.  I'm tracking that down before proceeding with framer stuff.

pkoning2:
1. Where is the indentation wrong?  I try to match the SIMH style, but it's hard to figure out what it actually is.
2. Ok.
3. The existing speed machinery is for UARTs, and accepts only UART speeds.  The framer accepts any speed between 500 and 1M, so it didn't seem to make sense to try to use the existing code.  I just copied the computation of the scheduling parameter from the speed.
4. Yes, I understand the concern.  The point is that there are differences between DMC (DDCMP 3.0) and DMR/DMP (DDCMP 4.0) mode that are visible in a sync framer, and from the documentation in the back of the DMP manual there are some that really matter.  So I implemented them as a mode setting in the framer.  At one point I had it specified at attach time, but with the DMP it's a choice specified by the driver when it starts DDCMP.   So the purpose of the start_framer call is to correspond to the DMC/DMP start operation.  I did find out in testing that DMP emulation doesn't actually work (for example, RSTS rejects the device outright).  So I suppose I could make it default to DMC mode and use another keyword in the attach string to say DDCMP 4.0 mode.  But either that would have to change for DMP, or be something it doesn't actually support.  And it gets worse if you wanted to emulate multipoint mode for DMP; the framer doesn't currently handle that but it could with enhanced firmware.

I've seen the same issue Reindert saw in the past, it appears if you stop the other end (drop it in to the sim> prompt) for long enough.  Apparently the timeout and retransmit keeps sucking up buffers if previous sends can't actually happen because the connection is blocked; I would think the solution is to detect that case and not queue up more stuff.

markpizz:
I'm thinking that you should include the framer device documentation right in the 
Simh PDP11 directory so all of the DDCMP stuff is in one place.  Migrate updates
there as you make evolving changes in the future.

In pdp11_dup.c, why_ did you make this addition to dup_detach:

> if (r == SCPE_OK)
>    r = tmxr_close_master (&dup_desc);

There may be many DUP devices, which share the dup_desc, closing one shouldn't close them all, right?

Your most recent commit removed the framer calls from pdp11_dup,c, but you left in a previously added line to pdp11_dup.c:

> dup_desc.dptr = dptr;

Your initial change to pdp11_dup.c also changed when the DUP device enables the TMXR receive enable bit (rcve) to be triggered on DTR going up rather than the original code which did this when RTS went up.  I can understand the theory, but was this a necessary change?  Meanwhile, it would seem that in addition to this logic, the downward transition of the appropriate signal should also disable the TMXR bit (and possibly reset the line state to be sure that any pending but currently buffered data is discarded).

pkoning2:
Framer documentation... I could include the framer API document, that would make sense.  But the whole set of design files is on Github, including schematics, firmware, and all that.  

The point of the close_master change was to close the ethernet that was opened by the framer attach, which is via open_master.  But confusingly, open_master also does per-line stuff and in fact for these devices that's the only kind that's ever used.  Is there a "close_line"?  That's what I really need here.

That dptr copy was to get trace messages from sim_tmxr to come out.  Without that change they never appear.  I can undo that change if you want.

The change to do receive enable via DTR is to allow the framer to be seen.  There was a deadlock: RTS is set if carrier is seen, but carrier can't be seen if you don't look at the framer to get status messages.  Apart from that, the logic is in the meaning of the signals.  DTR means data terminal ready, i.e., the device is ready to communicate with the modem (the framer, in this case).  RTS is request to SEND, and tying receive enable to a send request doesn't seem right at all.  Consider what would happen if someone were to run half duplex: RTS would be set when transmitting and clear when receiving, so when the emulated system wants to receive the simulator would be turning off receive.

markpizz:
> Framer documentation... I could include the framer API document, that would make sense. But the whole set of design files is on Github, including schematics, firmware, and all that.

Hmm..  What I meant was more like the framer marketing documentation...  OR, maybe the help could be expanded to cover that point.  I would definitely leave out any mention of Ethernet since that's an internal implementation detail a user needn't have any awareness of.  Likewise the pointer to your framer github stuff is all about how you implemented it which is good for you, but hardly required for someone who wants to use one (unless they've got to build it from scratch)...

> The point of the close_master change was to close the ethernet that was opened by the framer attach, which is via open_master. But confusingly, open_master also does per-line stuff and in fact for these devices that's the only kind that's ever used. Is there a "close_line"? That's what I really need here.

I'm thinking you need some framer code in _mux_detach_line.  Adding that there might change other stuff...

> That dptr copy was to get trace messages from sim_tmxr to come out. Without that change they never appear. I can undo that change if you want.

Hmmm... Let me mess with how the attach should be done so the device relationship is properly established.  Leave that there for now.

> The change to do receive enable via DTR is to allow the framer to be seen. There was a deadlock: RTS is set if carrier is seen, but carrier can't be seen if you don't look at the framer to get status messages. Apart from that, the logic is in the meaning of the signals. DTR means data terminal ready, i.e., the device is ready to communicate with the modem (the framer, in this case). RTS is request to SEND, and tying receive enable to a send request doesn't seem right at all. Consider what would happen if someone were to run half duplex: RTS would be set when transmitting and clear when receiving, so when the emulated system wants to receive the simulator would be turning off receive.

Good point.  I still think the DTR going down has the effect I mentioned above (turn off rcve and flush any pending input).

pkoning2:
Framer marketing documentation, um, yes.  I suppose I should see which item of the framer description is its "data sheet" which would serve that purpose.  https://github.com/pkoning2/ddcmp/blob/main/README.md is not quite it, I think I need to write something new.  Will do that, and then I'll drop it here as well.  So right in the PDP11 directory?  Is a .md file ok?

Ok, so I missed tmxr_detach_ln, I'll remove the close_master from pdp11_dup and put the removal of the framer state there rather than in close_master.

As for cleaning up receive stuff on DTR drop, that seems logical but it wasn't there before for RTS drop.  I can think about adding it but since that's an unrelated improvement I'd rather keep it as a separate work item.

markpizz:
>  Framer marketing documentation, um, yes. I suppose I should see which item of the framer description is its "data sheet" which would serve that purpose. https://github.com/pkoning2/ddcmp/blob/main/README.md is not quite it, I think I need to write something new. Will do that, and then I'll drop it here as well. So right in the PDP11 directory? Is a .md file ok?

A more thorough marketing like description in the help would be fine.  You'll also get a 1 line/paragraph mention in the simh/Readme.md.  If you want to drop a file in the PDP11 directory a .md file is great.

> Ok, so I missed tmxr_detach_ln, I'll remove the close_master from pdp11_dup and put the removal of the framer state there rather than in close_master.

That sounds right.

> As for cleaning up receive stuff on DTR drop, that seems logical but it wasn't there before for RTS drop. I can think about adding it but since that's an unrelated improvement I'd rather keep it as a separate work item.

Fine.  I'll deal with that detail along with getting the attach done right.

markpizz:
> > That dptr copy was to get trace messages from sim_tmxr to come out. Without that change they never appear. I can undo that change if you want.

Actually, that line you added is redundant.  dup_desc.dptr is set in dup_reset.

pkoning2:
BTW, just to clarify: as things stand right now, if you want a framer you'd have to build it from the open source design files.  I'm not offering boards (bare or assembled).

R-Voorhorst:
Nice, notice BTW in Dutch :Belasting :) Toegevoegde Waarde of which VAT only gives a subset of some possible meanings …

 

 

Reindert

markpizz:
Although your engineering development name for this stuff is FRAMER, maybe the right name for the user facing part within TMXR is SYNC or something like that.  Thoughts?

Maybe we also add a SHOW SYNC command which lists the available SYNC devices like the SHOW ETHERNET or SHOW SERIAL does now and the attach syntax then refers to a SYNCn name displayed in the SHOW SYNC command.  Like the ethernet and serial cases, an attach can either use the SYNCn name or whatever the SHOW SYNC eth name that was displayed.  This (and the whole framer operation) presumes you can actually identify a framer eth device from other network interfaces...

Please enrich the help to explain more about what framer mode "INTERNAL" means which should probably include examples.

pkoning2:
"SYNC" instead of "FRAMER", sure, that seems like a reasonable change.  I called it "framer" because it isn't just a raw sync line interface (like, say, a DU-11) but it also includes the DDCMP frame delineation function.  That fact really only matters to someone using the API, though.

As for "show sync" and "show ethernet", it certainly is straightforward to distinguish a framer provided the ethernet interface MAC address is available at the time the show command is run.  The framer uses MAC addresses starting with aa-00-03, which isn't a value in use anywhere else.  It was assigned to DEC but never used, and later on it became obsolete as part of the 802.1 addressing rules.  So I grabbed that OUI.  :-)

pkoning2:
Forgot to talk about "internal".  Looking back I think that's a naming error I made.  DEC refers to the coax-based connections as "integral modem" (also "local").  "Integral" because the modem function is built into the line card, rather than provided by an external telco-supplied box as would be the case for RS232 or V.35 or the like.  I included "coax" as an alternate keyword; should I make that be the primary one?  (And fix the other to be "integral"?)

markpizz:
I don't have enough direct experience in this domain to have a preference or useful knowledge about the terms you use.  My concern is that whatever terms you use, they're explained so that an interested, but not crazily involved user can understand what they mean and how it might be relevant to various use cases....

markpizz:
Looking at your most recent commit:
- When I suggested 
> Maybe we also add a SHOW SYNC command which lists the available SYNC devices like the SHOW ETHERNET or SHOW SERIAL does now and the attach syntax then refers to a SYNCn name displayed in the SHOW SYNC command. Like the ethernet and serial cases, an attach can either use the SYNCn name or whatever the SHOW SYNC eth name that was displayed. This (and the whole framer operation) presumes you can actually identify a framer eth device from other network interfaces...

What I meant was, like the serial and ethernet cases SCP presents to a user ethN and serN names which users can specifically attach to, OR they can use the detailed (expanded) OS device name displayed by SHOW ETHERNET or SHOW SERIAL for each listed ethN or serN.  These names are abstracted to hopefully make the SCP internal state (contained in a SAVE/RESTORE) theoretically portable to other similar or completely different host systems.  In the case of ethernet, the value is minimally useful since different systems almost always will have different sets of network interfaces.  In the serial case, ser0 might be one specifically dedicated to simulator use and thus more commonly be correct.  In the sync case, it would seem that sync0 would always actually be correct across hosts since these devices are rare and only a dedicated user would have them on their system.

Although all the current use cases of your sync stuff are DDCMP devices, I've gotten the impression from many things you said, that your device is a general purpose sync device.  As such, I'd remove specific mention of DDCMP from the what is displayed by SCP relating to this unless it is part of the output for a specific device SHOW SYNC that is in a DDCMP mode, or some parameter to attach which specifies DDCMP.

The show_sync logic should exist in tmxr and tmxr_show_sync_devices should be there along with the MTAB display support routine tmxr_show_sync.  No need for singleton extern declarations in pdp11_dup & dmc.

In sim_ether _eth_devices is getting the host nic address (this is fine to properly determine framer devices and store the description, but you are saving the hwaddr which isn't needed at all.  Yes, you later reference the hwaddr in eth_show_fr, but that check would produce equivalent results if the description was compared to "DDCMP synchronous interface" or merely "synchonous interface" as suggested above.

pkoning2:
I see what you're getting at with "show sync".  The difficulty is that, under the covers, they ARE ethernet devices.  I suppose I could make up "sync<n>" aliases which map to the actual Ethernet interface name, so you'd see something like: "sync0  en11" rather than the current "eth3 en11".  I don't know what is involved in making up those "sync" pseudo-device names, though.
On keeping the hwaddr, I was thinking of having "show eth" show it, but it doesn't really given the width of the pseudo-device column.  So I can take it out, the most obvious alternative is to replace it by a true/false flag.  (Checking for a literal string seems a bit hacky.)
On the framer capabilities, perhaps I have not been very clear.  The current device is very definitely DDCMP specific; it does packet framing and that framing is specific to DDCMP.  The underlying hardware can do other sync protocols easily enough, but that would require substantially different firmware.  I've thought a bit about HDLC, and a bit less about Bisync.   But I have no personal experience with those, so the incentive for me to get into those options isn't really there.  (If someone else wants to expand the firmware to cover these, I'd certainly help with suggestions and the like.)  Also, the coax interface is designed to be compatible with the DMC integral modem line card.  It's possible that it could drive other coax-connected devices -- for all I know the hardware could, say, generate IBM style coax HDLC signals.  But that's not something I set out to do and if it's workable at all it too would take more development and some luck as well.  Last but not least, the device API is DDCMP specific; it sends and receives DDCMP frames to/from the SIMH host.

pkoning2:
Ok, I think I see how to make  "sync<n>" pseudo-devices.  Will take a look at doing that.  Probably a couple of days before I have time.

markpizz:
OK.  Please keep the primary sync logic in TMXR with the framer specific support in sim_ether.  

You (or someone) may come along with a framer variant for other than DDCMP at some time which will then extend in sim_ether logic for that.  Meanwhile, someone might be motivated to come at the sync problem with other hardware which isn't framer based and that would then dispatch from TMXR.

Thanks.

pkoning2:
So, put the "sync<n> name handling in sim_tmxr?  
I noticed that the eth<n> check in eth_open is broken: it assumes that <n> is a single digit which is not necessarily the case.  On my laptop, because of various OS-generated interfaces, I see eth0 through eth13.  Should I fix that?

markpizz:
Putting the sync name stuff in tmxr is the appropriate place since any sync capable simulator device will automatically be using tmxr anyway.

Yes, fix the ethN number size assumption.

pkoning2:
The question is how to make a clean connection from sim_ether to sim_tmxr.  The routine used by "show sync" is _eth_devices which is internal to sim_ether.  
One possible answer is to make that a public interface, call it from sim_tmxr, and have sim_tmxr pick up just the entries in the returned list that are sync devices.  I can see some other options but none of them feel better; in every case some Ethernet internals are necessarily exported to sim_tmxr, or ddcmp device knowledge imported into sim_ether, because that's the nature of the beast.

markpizz:
> I can see some other options but none of them feel better; in every case some Ethernet internals are necessarily exported to sim_tmxr, or ddcmp device knowledge imported into sim_ether, because that's the nature of the beast.

sim_ether functionality is already (and reasonably) leveraged by the tmxr sync support since the needed access is via ethernet.

Thinking a little more here and I realize that SHOW ETHERNET really shouldn't list the framer devices as an ethN item, which means that _eth_devices should be modified to remove them.  

There should be a public routine exported from sim_ether:
int eth_framer_devices (int max, ETH_LIST *list)
which returns the list of framer devices.  This routine would be cloned from _eth_devices() and only return the framer DDCMP or other protocol framer devices.

TMXR can then leverage this along with minor variation of the logic in sim_serial.c to support the needed syncN semantics.  Internal to TMXR there should be sync_read and sync_write which now merely call the related eth routines and in the future might support either eth based dispatch or another flavor if an alternate sync paradigm exists.

markpizz:
Merged manually.  Thanks.

PR #1110, opened 2022-01-06
Fix crash when using EP multiply

k1mu:
When performing floating point multiplication, the code overwrote an addtional word of the floating point fraction with zeroes. This is harmless with standard FP, as the floating variables always have space for EP-length vars. When doing an EP multiply, this causes a word on the stack to be zeroed. For the latest Raspbian release, this causes a segfault as there's no padding past that var on the stack.

This fix, which has been tested against the original crashing code plus validated using the FPP-8 diagnostics, avoids the overwrite.

### DIFF
--- a/PDP8/pdp8_fpp.c
+++ b/PDP8/pdp8_fpp.c
@@ -25,6 +25,8 @@
 
    fpp          FPP8A floating point processor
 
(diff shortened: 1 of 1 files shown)


markpizz:
Merged with cherry pick reformatting the commit message.

Thanks,

PR #1120, opened 2022-02-02
PDP11: Fix bug in KG11 emulation

pkoning2:


markpizz:
Merged with Cherry Pick

PR #1131, opened 2022-03-08
AltairZ80: Resolve CID 1514796,1514860, 1514861.

hharte:
FYI, @psco


psco:
Looks good, thanks Howard

PR #1134, opened 2022-03-11
TT2500, IMLAC: Fix two Coverity issues.

larsbrinkhoff:
Coverity found three issues.  This fixes two of them.

### DIFF
--- a/imlac/imlac_pt.c
+++ b/imlac/imlac_pt.c
@@ -144,7 +144,7 @@ ptp_iot (uint16 insn, uint16 AC)
   if ((insn & 0771) == 0271) { /* PUN */
     ;
   }
-  if ((insn & 0772) == 0274) { /* PSF */
+  if ((insn & 0774) == 0274) { /* PSF */
     ;
   }
   return SCPE_OK;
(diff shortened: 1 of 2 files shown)


markpizz:
Actually Coverity found 5 tt2500 and 1 imlac.

Use the scroll bar below the issue list and scroll all the way over to the File column.  Clock on the column header and the issues will be grouped by the directory/file names and then they will all be visible together.  Note also that there currently are 2 pages of simh issues.

larsbrinkhoff:
I'm aware there are more issues, but I don't have an immediate solution for the others (which is one problem manifesting as several).

However when I went to refresh the view this morning, it had changed.  It now lists "Component" instead of "File" and there's nothing to match tt2500.  I don't see anything in "My Outstanding" list either.

larsbrinkhoff:
Never mind, I see them again.

larsbrinkhoff:
As far as I can see, the four remaining tt2500_cpu.c issues are false positives, due to Coverity mistakenly using the ALTAIR/altair_sys.c instead of tt2500_sys.  The latter file implements fprint_sym in a way that does not use `val` as an array indexed beoynd 0.

Screenshot from 2022-03-12 17-21-23

markpizz:
Indeed.  All of the fprint_sym() routines have the same signature and the potential semantics of that signature allow for *val to be used as an array.  Mark these as “False Positive” “Ignore” with a comment that says "the only use in this simulator is as a singleton".

PR #1152, opened 2022-04-21
Fix DUP11 for DDCMP CRC perf. option and enable DUP11 on 8600

trevorwarwick:
This change is necessary to make the DECnet/VAX Phase V DUP11 driver work in the simh environment. This VMS driver implements the CRC performance option in section 5.1.2 of the DDCMP spec., and thus relies on receiving a valid character after the CRC on a received frame.  I refined my original fix after some discussion with Paul Koning, and it's done in the way that he suggested would have least possible impact on any other driver.    I've also added the DUP11 to the default device list in the 8600 simulator as it was missing.


markpizz:
Hi Trevor,

I'll get to this in the next few days.  

Meanwhile, can you confirm the combination of endpoints you've tested with this change?

Specifically, DUP11<->DMC11  DUP11<->KDP11, etc. varying the host simulators/operating systems in the configs being tested...

I find it surprising that DECnet Phase V had DUP support while otherwise no such device has historically been in use on VAXen.  This seems very late in the day....

- Mark

trevorwarwick:
So far, I haven't tested anything else. I don't think it's interoperability with other devices that's relevant for this change, it's with host drivers only.  I'm not sure what other systems support DUP11/DDCMP but at the very least I can ask Paul K and Johnny B if they can try it out on RSTS and RSX.  If those work I think it would give good confidence against regressions.

trevorwarwick:
I'm pursuing getting some testing on other OSes. 

In the meantime, I agree, it's surprising that we supported the DUP/DPV11 with DECnet-VAX Phase V, since the main focus was on the more modern devices with higher line speeds like DSV11/DSB32/DSH32.  I think it was a combination of things: DUP/DPV were previously supported for VMS SNA and VAX PSI, but each layered product used a different driver. The Phase V implementation was very modular, so it wasn't that much work to add DDCMP support to a driver that was going to do SDLC/HDLC anyway. Also we obsoleted the microcoded DDCMP devices (DMC/DMR/DMV), so perhaps the idea was to keep another option open if customers didn't want to buy a more modern device.

Anyway, the upshot is that the DUP11 is actually the only serial device in simh that can work with DECnet-VAX Phase V, so I think it's important to get it working properly.

markpizz:
Given this extra utility, maybe you want to take a crack at completing the DPV11 functionality since by the time Phase V came around, I think the vast majority of useful systems were Qbus...  

I never had a good setup to test the DPV since my world experience was VMS and nothing on the PDP11.  By the time Phase V came along I was on to other things and never set it up.

PR #1155, opened 2022-04-29
sim_imd: Check for EOF when reading the IMD header

agn453:
The recent change to check that the IMD file track header is 5 bytes was attempting to read beyond the end-of-file, resulting in a spurious file open error and (under AltairZ80) a I8272: IMD disk corrupt diagnostic.

This change exits the IMD file parse if the sim_fread returns 0 bytes and the end-of-file is reached.

### DIFF
--- a/sim_imd.c
+++ b/sim_imd.c
@@ -119,7 +119,7 @@ static t_stat diskParse(DISK_INFO *myDisk, uint32 isVerbose)
     uint8 sectorHeadMap[256];
     uint8 sectorCylMap[256];
     uint32 sectorSize, sectorHeadwithFlags, sectRecordType;
-    uint32 i;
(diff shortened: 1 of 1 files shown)


markpizz:
@psco and @hharte please review this change.

Once @psco and @hharte are good with this, @agn453, will you redo this commit with the above explanation in the commit message.  This lets the repo stand own for walking back in history (via blame or otherwise) rather than having to chase all over to understand each change.

Unwind the commit locally, commit again with revised commit message and push -f back to your github account.

Thanks.

hharte:
The change looks good, thanks @agn453 .

psco:
Also ok from my side. Thanks to @agn453 for finding and fixing this defect.

PR #1160, opened 2022-05-18
PDP11, Qbus & Unibus VAX: Fix DUP device loopback on device reset

trevorwarwick:
- When the DUP TXCSR DEVICE_RESET bit is set, which clears the MAISEL bits, any
existing tmxr loopback must also be removed.


trevorwarwick:
The VMW Phase V driver code has a test mode that puts the device into loopback.   The internal loopback within the tmxr layer wasn't being cleared on device reset, so once the device was put into loopback, it could never be successfully used for normal communications without a reboot.

PR #1167, opened 2023-01-04
SEL32: Add IPU device support using pthreads on Linux and Windows.

AZBevier:
SEL32: Update makefile and SEL32.vsproj file to add IPU device.
SEL32: Update README.md file.
SEL32: Do a general code cleanup.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,7 +1,7 @@
 image:
-  - Visual Studio 2015
-  - Ubuntu2004
   - macos-monterey
+  - Ubuntu2004
+  - Visual Studio 2015
 
 platform:
   - x86
@@ -58,7 +58,7 @@ on_success:
   - ps: |
         $id_line = Select-String -Path $commit_id_file -Pattern 'SIM_GIT_COMMIT_ID'
         $id = $id_line.Line.SubString($id_line.line.IndexOf("SIM_GIT_COMMIT_ID") + 18, 8)
-        $datetime = Get-Date -Format "yyyy-mm-dd_hh-mm-ss"
+        $datetime = Get-Date -Format "yyyy-MM-dd_HH-mm-ss"
         $version_major = (Select-String -Path 'sim_rev.h' -Pattern '#define\s+SIM_MAJOR\s+(\d+)').Matches.Groups[1].Value
         $version_minor = (Select-String -Path 'sim_rev.h' -Pattern '#define\s+SIM_MINOR\s+(\d+)').Matches.Groups[1].Value
         $version_mode =  (Select-String -Path 'sim_rev.h' -Pattern '#define\s+SIM_VERSION_MODE\s+"(.+?)"').Matches.Groups[1].Value
@@ -75,7 +75,7 @@ on_success:
           Compress-Archive -Path ..\BIN\NT\Win32-Release\*.exe -DestinationPath "$pkg_filename.zip"
         } else {
           pushd ../BIN
-          tar -cvzf ../Development-Binaries/$pkg_filename.tgz --exclude=buildtools '--exclude=*-build' ../BIN *
+          tar -cvzf ../Development-Binaries/$pkg_filename.tgz --exclude=buildtools '--exclude=*-build' *
           popd
         }
         "$platform Build results for simh Commit: https://github.com/simh/simh/commit/$id" | Out-File -FilePath ../pkg_commit_message -Encoding ascii


markpizz:
Hi Jim,

I wish you would have discussed this before going down the implementation path.

There is a specific rule that simulator code not call any OS library libraries directly.  Any interface to OS capabilities (threads in this case) needs to be implemented directly in a SCP provided library.

From a 3 second glance at the code in this PR I'm guessing that you're adding support for multiple CPUs to your simulator.  Doing this in threads is complex and unreliable in various ways depending on host platform.  The only universally supported simh model for multiple CPU support is for the sim_instr() API to manage CPU specific context separately as needed according to the specific needs of the architecture being simulated and essentially to round-robin instruction execution between each of the potentially many simulated CPUs in sequence...  There is at least one implementation of this in the current simulators, that would be the B5500 CPU.

- Mark

AZBevier:
Hi Mark,

I was not aware of any specific rule for using pthreads.  They are in 
the simh release so I figured they were there to use.  The SEL32 IPU is 
a compute only device that does not handle any interrupts or I/O.  Those 
instructions are trapped and sent back to the CPU for execution.  One 
mutex is used to signal back and forth between the IPU and CPU.  UTX and 
MPX-32 software handles the IPU without any issues.  I was the original 
software engineer back in early 80's that developed the MPX-32 IPU 
software.  It is really straight forward is not "complex" at all.  We 
have tried two other methods which work even faster and better than 
pthreads, but Windows does not support the fork() from UNIX.  Let me 
know if you have any issues.

Jim

rcornwell:
Mark,

  I would not use the B5500 as an example of how to do support for multiple CPU's. While the second CPU was similar to the IPU on the SEL32, it is also different. On the B5500 the two CPU's were pretty much the standard configuration. Also the B5500 second CPU kind of needs to be in sync with the master CPU. The B5500 was more of a master slave configuration where the Sel32 is IPU is more an auxiliary processing engine,

markpizz:
The IPU shares memory with the primary CPU, right?

AZBevier:
The IPU shares all of memory with the CPU, but has all of it's own 
registers.configuration memory.  Trap vectors are assigned different 
memory locations for IPU/CPU.

Jim

markpizz:
Consistency in memory access is the potentially complex and unreliable host platform specific concerns.  

The IPU shares all of memory with the CPU, but has all of it's own registers.configuration memory.  Trap vectors are assigned different memory locations for IPU/CPU.  I suspect that this part is parallel to the B5500 situation.  

The IPU could be a DEVICE which executes n instruction (possibly 1) at a time and is scheduled (via sim_activate) to run every n(or m) instructions when it is supposed to be active.  There are coprocessor devices in at least one simulator that run like that.

AZBevier:
The only instructions that need a read/lock feature are the set bit in memory (SBM) and the zero bit in memory (ZBM) and these are protected with a pthread mutex.  That is, lock, r/w, unlock.  Those are the only instructions that the original hardware makes r/w with lock when system has an IPU configured.  Otherwise, r/w access was/is not an issue. There is really no reason to have simh interact with the instruction stream because only instructions are being executed that do no I/O or interrupt processing that requires simh.  Just parallel instruction execution.  That was the intent of the IPU by SEL.

geertrtew:
> The only instructions that need a read/lock feature are the set bit in memory (SBM) and the zero bit in memory (ZBM) and these are protected with a pthread mutex. That is, lock, r/w, unlock.

The alternate configuration uses fork(), Posix Semaphores and mmap to share the memory to achieve the same. That works best under Unix/Linux and pthreads/mutex work best under MS Windows. Assuming you have multiple cores in your host, you really get double power: run a single CPU bound program in the same time as a double one. Obviously, you can only use fork() here when not using other peripherals in the IPU. It is a slave processor as in asymmetrical multiprocessing

I had a severe cultural chock during the project: on a PDP-11 the HALT instruction makes your PDP dead as a monkey and on a SEL32 it will wait for a trap silently. ;-)

rcornwell:
I would go with Pthread over Fork and shared memory. This is more portable.

On PDP11 Halt means stop and do nothing. There is a WAIT instruction which does the what you want.

geertrtew:
> I would go with Pthread over Fork and shared memory. This is more portable.

Yes, it is more portable and it will enable implementations of 2nd 3rd etc CPUs that share peripherals. Nevertheless pthreads give funny effects under Unix/Linux. Unbalanced performance!!
> 
> On PDP11 Halt means stop and do nothing. There is a WAIT instruction which does the what you want.
DYK both HALT and WAIT as binary compatible beween PDP11 and SEL32, altough the bit in WAIT is numbered 0 in PDP11 and 31 in SEL32 ;-)  Ohhh, WAIT in SEL32 traps when ints are disabled and your PDP will hang when PIO=7 in the PSW.

rcornwell:
I am pretty sure the PDP11 behaves correctly.  HALT gives control back to console, until something on front panel tells it to continue it will stay at console. WAIT hangs the processor until an interrupt occurs. If Priority is 7 then no interrupts can occur, so the machine would be hung. I would say that SEL32 is broken. If interrupts are disabled it should never take an interrupt! :-).

geertrtew:
> I would say that SEL32 is broken. If interrupts are disabled it should never take an interrupt! :-).

When setting the very first baby steps with the CPU/IPU in my Lab, I correctly assumed that UTX was ready to go for it and complained to Jim that the IPU hopped into a HALT on each attempt. The chock came later....

PR #1174, opened 2023-05-10
sigma: implement WAIT instruction

kenrector:
Added code to implement WAIT operation in sigma_cpu.c and correct the register returned for SIO, TIO, and other io instructions


markpizz:
Merged a while back.

PR #1181, opened 2023-08-23
BESM6: Corrected handling disk device groups.

leobru:
Device groups now work; up to 64 disk drives may be used.


leobru:
AppVeyor got confused.

PR #1186, opened 2023-09-30
correct logic error in build_vstudio.bat for VS2022 and v141_xp

jdbruner:
``build_vstudio.bat`` converts the ``.vcproj`` files from Visual Studio 2008 to ``.vcxproj`` when used with newer versions of Visual Studio and configures it to build XP-compatible binaries using the ``v141_xp`` toolset. However, this toolset is deprecated and may not be present when using Visual Studio 2022. The script intends to edit the ``.vcxproj`` files only in the case where it exists; however, due to a logic error it actually does this unconditionally. This causes the build to fail on VS2022 when the XP toolset is not present.


markpizz:
Hi John,

This has been merged with you as author and your PR comments in the commit message.

Thanks,

- Mark

BTW: The build doesn't complete successfully for all simulators with VS2022 since, for some reason, Microsoft has chosen to make *some* static libraries incompatible when they are compiled with one intermediate version of VS2022 vs subsequent versions.  Complete functionality and binary compatibility back to XP is available when compiled with VS2008.  Instructions about where to find VS2008 are available in the "Visual Studio Projects\0ReadMe_Projects,txt" file.

PR #1188, opened 2024-01-13
SCP: Adjust timer calibrations if SET THROTTLE x/t type is enabled

deltecent:
SCP does not adjust the timer calibrations when `SET THROTTLE x/t` is enabled causing the timers to run slow.

This PR fixes that problem so simulators run at the correct speed when execution begins.


markpizz:
Patrick,

Thanks for coming with this.  The whole fix for this and other things is included in another change I'm working through.  The throttle issue will be fixed when that is ready.  Good digging to fix this specific problem.

- Mark

PR #1192, opened 2024-02-21
sigma: Fix three kinds of errors in device modules.

kenrector:
Fix three kinds of errors in device modules.  Discovered while getting standalone System Exerciser to run.
IO: DVT_NOTDEV macro incorrect, Device mapping algorithm creates false dispatch points.
This mapped Multi Unit Controller and Single Unit Controller to same device. 
DP, DP, MT, RAD:  Test for non-existent device returns wrong status.
DP, DK, MT: TIO status should return non-operational for unattached device.


markpizz:
Merged

PR #1193, opened 2024-03-28
sigma: Add chaining modifier flag to io

kenrector:
sigma: Implement CM (Chaining Modifier) flag to IOP simulation in sigma_io.c

### DIFF
--- a/sigma/sigma_io.c
+++ b/sigma/sigma_io.c
@@ -363,6 +363,7 @@ if (CC & cpu_tab[cpu_model].iocc)                       /* error? */
 chan[ch].chf[dev] = 0;                                  /* clear flags */
 chan[ch].chi[dev] = 0;                                  /* clear intrs */
 chan[ch].chsf[dev] |= CHSF_ACT;                         /* set chan active */
+chan[ch].chsf[dev] &= ~CHSF_CM;                         /* clear chain mod */
 chan_new_cmd (ch, dev, R[0]);                           /* new command */
 return st;
 }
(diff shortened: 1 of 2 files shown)


markpizz:
Ken,

The convention is to have your commits use the format previously described where the affected simulator is present at the beginning the commit message.

In this case, it should say:
"sigma: Add chaining modifier flag to IO" or "sigma: implement CM (Chaining Modifier) flag to IOP simulation"

This way, the change is precisely visible in history that is part of EVERY copy of the repository.  The PR details are a web based add on which absolutely is not needed for someone to use the repository.

You can fix this locally without jumping through hoops canceling the PR merely with the following commands:

    $ git reset --soft HEAD^
    $ git commit using the appropriate commit
    $ git push -f

\- Mark

kenrector:
Mark,

should that be    git push -f origin chan_modifier

?

kenrector:
Mark,

Sorry, I never seem to get this right.  
The simulator name should be in the commit message, not the pull request note.

Wouldn't the simulator name be kind of redundant since the commit belongs to the
simulator?

markpizz:
@kenrector said:
> should that be git push -f origin chan_modifier

Well, that depends on how you've got your git remote's setup.  If "git push origin chan_modifier" was how you pushed the "chan_modifier" branch to your github repo, then adding the -f will replace the current branch state that is referenced in the PR.

> The simulator name should be in the commit message, not the pull request note.

The simulator name (or otherwise affected component: (SCP, ETHER, SOCK, VIDEO, etc.)) should be the first part of the commit message.  That first message line should be <= 80 characters.  If more commit comments are appropriate for any particular commit, then the first line should be as described followed by a blank line and then addition commit comments wrapped < 80.

The simulator name CAN be part of the PR title, but this discussion is about the commit messages.

> Wouldn't the simulator name be kind of redundant since the commit belongs to the
simulator?

Any git commit belongs to a particular spot in a branch in the repository and, in the simh repo has nothing which knows about simulators.  There are some 78+ simulators in the simh repository.  The point of having the simulator name start the commit messages makes it easy to see which simulator (or component) is affected by particular commits when glancing through the brief form of the commit history log.  You can see this log by:

    $ git log
or

    $ git log | grep -i sigma

PR #1196, opened 2024-06-01
AltairZ80: Changes to JAIR devices

deltecent:
This PR makes the following changes to the JAIR devices:

Refactors the service routines into STAT, TX and RX units (thanks @markpizz)
Adds a 128-byte input ring buffer
Corrects a null pointer exception
Adds overrun status bit implementation
Replaces a tab with spaces


markpizz:
Merged - Thanks

PR #1201, opened 2024-09-03
AltairZ80: Adds "ROM" option to "load -h"

deltecent:
Adds the ability to mark pages as ROM when loading Intel HEX files.


deltecent:
@psco discovered that this PR does not work correctly with extended memory. I will be pushing another PR containing his fix.

PR #1219, opened 2025-08-15
Reverted tape mark behaviour as revision was wrong (Van Snyder)

Prince-Stolas:
I talked to Van Snyder and he told me about this "fix" he wanted introduced in SimH i1401. The fix was needed to get a program running. The program not working was likely due to a transcription error, not the emulator. This "fix" thus introduced a bug, which this pull request will fix.

Currently SimH i1401 doesn't deposit a group mark in core memory after reading a tape mark. It only deposits the tape mark. This is wrong. The tape mark should be followed by a group mark.

### DIFF
--- a/I1401/i1401_mt.c
+++ b/I1401/i1401_mt.c
@@ -25,6 +25,7 @@
 
    mt           7-track magtape
 
+   15-Aug-25    VRS     Reverted tape mark behaviour as revision was wrong (Van Snyder)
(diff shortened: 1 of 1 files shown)


markpizz:
I rewrote the commit message to be in the simh project standard format and included the above PR description.
The result, with you as the author, has been merged into the master branch

PR #1221, opened 2025-08-26
AltairZ80: Adds banked support for ExpandoRAM

deltecent:
SD Systems ExpandoRAM bank is selected through I/O Port 0xFF. This is required to run SD Systems banked CP/M 3.0.

Adds ability to override start address when loading HEX files. This is useful when loading HEX files for ROMs which start at address at 0x0000 that needs to loaded in the ROMs address space such as 0xF800.


markpizz:
Is this change good with @psco?

PR #1232, opened 2025-11-15
Altair8800: New Altair 8800 (BUS) Simulator

deltecent:
This PR is the first public release for peer review of a new Altair 8800 SIMH simulator.

Why another Altair simulator? AltairZ80 has been described as a “software simulator”, where the intent is run software designed specifically for executing under a simulator. Altair8800 is intended to accurately simulate the Altair hardware and execute software that will run unchanged on real hardware. Software and disk images can be moved between the Altair8800 simulator and real Altair and other S-100 hardware without any changes. To accomplish this, the following are major differences between AltairZ80 and Altair8800:

* The monolithic design where devices access other devices directly through external variables and functions is no longer supported. All devices exchange data through a new BUS device. Memory and I/O address decoding and transfers are now handled by the BUS device. All interrupt requests are handled by the BUS device. "SHOW BUS VER" will show the configuration of the BUS.

* System RAM was moved from the CPU device to a new RAM device and managed by the BUS device.

* Banked RAM was moved from the CPU device to a new BRAM device. Banked RAM can only be accessed through the BUS device. Memory in banks that are not currently selected cannot be accessed. The AZ80 “banked” RAM was removed.

* ROMs were moved from the CPU and DSK devices to the new ROM device. The original MITS Disk Boot Loader, Turnkey Monitor, and Hard Disk Loader are available. Mike Douglas’ Altmon Monitor and AltairZ80 ALTAIRROM are also available.

* The original MITS Disk Boot Loader is now the default boot loader.

* The monolithic Multiple-CPU/RAM/ROM/IO/BankedRAM CPU device has been replaced with a generic CPU device that provides an abstraction layer between SIMH and the supported CPU architectures (currently 8080 and Z80). All IO is handled through the BUS device. RAM, Banked RAM, and ROM are each handled by their own independent devices.

* The AltairZ80 SIO device was replaced with the M2SIO0 and M2SIO1 devices. The M2SIO devices fully support TMXR.

* A new SIO device was added to provide generic, programmable, Serial IO. TMXR is not supported on this device.

* The Altair 8800 did not have PTR or PTP hardware devices and have been removed. PTR and PTP software devices are defined by the software executing on the simulator.

* Port 0xFF sense switches was moved to a new FP device and programmed output was moved to a new PO device.

* The SIMH pseudo device no longer uses the removed PTR and PTP devices. The SIMH device has its own IO system. To avoid conflicts with other devices and remain compatible with the R and W utilities written for AltairZ80, SIMH “borrows” I/O ports 12H and 13H during file transfers. Only SIMH commands needed to support R and W file transfers are supported. All other SIMH commands were removed. AltairZ80-specific versions of CP/M are not supported by Altair8800.

* PC queue was removed from the CPU device.

* The Altair8800 simulator only supports the Altair 16-bit address and 8-bit data buses. 8086 and 68K CPU architectures were removed.

* All CPU timing (clockFrequency) and “sleeps” (SIO SLEEP) have been removed. SIMH THROTTLE is fully supported and is the recommended way to manage simulator speed and host CPU utilization. Executing “SET THROTTLE 100K/1”, for example, should provide ample speed without tasking the host CPU.

* HEXLOAD and HEXSAVE commands were added. The LOAD “-h” option has been removed. Intel Hex and sRecord (coming soon) formats are supported.

* A new DSK library was added to provide a standard method of managing disk images of various soft sector formats.

* The WD179X device was replaced with the WD17XX library. The WD17XX library uses the DSK library for all disk image manipulation. Support for the proprietary IMD disk image format was removed.

* The TARBELL floppy disk controller device was refactored to use the new DSK and WD17XX libraries. Other AltairZ80 floppy disk controllers will be refactored and added to the Altair8800 simulator.


deltecent:
This PR has been replaced by #1234

PR #1243, opened 2026-03-01
fix null pointer dereference when SHOWing cpu register

jdbruner:
When doing a SHOW of a CPU register, ``show_cmd_fi`` sets ``dptr`` to ``sim_dflt_dev`` (the CPU) but doesn't set ``uptr``. Previously this was OK; however, there is new code that that depends upon ``uptr`` being non-NULL. As an example, doing a ``show history`` in a PDP11 simulator will crash with a segmentation violation.

A second, less serious, problem is that ``sim_snprint_sym`` doesn't null-terminate the buffer it passes to ``strlcpy``. ``strlcpy`` expects the source string to be null terminated. In this case, the destination size is always less than or equal to the length of the source string, so the result is well-formed (and null-terminated). However, ``strlcpy`` will look for the end of the source string in order to return its length. ``sim_snprint_sym`` doesn't depend upon this, so usually this is not a problem. However, when ``strlcpy`` goes past the end of the source it could land in unallocated memory. (As an example, this happens when simh is run on Windows with app verifier.) The fix is to ensure that there is always a null byte at the end of the source string.

### DIFF
--- a/scp.c
+++ b/scp.c
@@ -2977,6 +2977,7 @@ if (vm_flag || ((reason = fprint_sym (sim_tmpfile, addr, val, uptr, sw)) > 0)) {
     reason = dfltinc;
     }
 sim_mfile = NULL;
+mbuf.buf[MIN(mbuf.pos,511)] = '\0'; /* ensure null termination */
 strlcpy (buf, mbuf.buf, MIN(bufsize, mbuf.pos + 1));
 free (mbuf.buf);
 return reason;
@@ -17989,3 +17990,4 @@ for (i = 0; (dptr = sim_devices[i]) != NULL; i++) {
     }
 return stat;
 }
+
--- a/scp.c
+++ b/scp.c
@@ -6777,6 +6777,7 @@ else {
                          (mptr->pstring && (MATCH_CMD (gbuf, mptr->pstring) == 0))) ||
                         (!(mptr->mask & MTAB_VDV) && (mptr->mstring && (MATCH_CMD (gbuf, mptr->mstring) == 0)))) {
                         dptr = sim_dflt_dev;
+                        uptr = dptr->units;
                         lvl = MTAB_VDV;                 /* device match */
                         cptr = svptr;
                         while (sim_isspace(*cptr))


markpizz:
I've added your null pointer fix to pending changes which will be pushed out soon.

The NUL termination logic is concerning for the following reasons:
    - As you say, the size argument to strlcpy() specifically limits the copy buffer size to the smaller of the actual buffer size and the amount of data that was produced + 1
    - Your logic will remove the last byte of the data that has been produced.

You say: "strlcpy expects the source string to be null terminated."  Actually, it would be "nice" if the input string were NUL terminated, BUT it is defined to always NUL terminate the copy result which the existing logic counts on.

passing the smaller of the buffer size of the size of the data that was produced + 1, if strlcpy() is well behaved, will do the right thing and always produce a null terminated string with the desired data.

jdbruner:
This doesn't affect the content, because the last content byte is at offset mbuf.pos-1.

The last byte of the source isn't copied into the destination, because the size is exhausted and sim_strlcpy ensures that the result is null-terminated.

The problem is that once it has filled the full size of the destination buffer, sim_strlcpy traverses the rest of the source buffer to determine its length:

```c
    /* Not enough room in dst, add NUL and traverse rest of src */
    if (n == 0) {
        if (size != 0)  
            *d = '\0';              /* NUL-terminate dst */
        while (*s++)
            ;   
        }
return (s - src - 1);               /* count does not include NUL */
```

This depends upon the source being null terminated, but it's not, so it is indeterminate how far the ``while(*s++)`` will go. Eventually there will be a null byte. That means the returned count could be anything, but fortunately the returned value isn't used However, if heap memory isn't contiguous, then the ``while(*s++)`` could hit an unallocated page and get a segmentation violation.

This happened to me when I tried running simh with app verifier in Windows (when I was investigating something else). App verifier deliberately puts heap allocations in isolated pages (initialized with non-zero bytes) to find things like this. Frankly, the risk of it happening under other circumstances seems pretty low, so I'm OK to drop it.

markpizz:
OK.  You're right.  

Thanks for dragging me through the problem.  I hadn't considered the return value of strlcpy().  

Like the other change, this will be amongst the other changes which will be pushed out soon.

jdbruner:
A modified version of this change (including something I missed) is in commit c6ec912, so I'm closing this PR.

PR #1247, opened 2026-04-21
Fix minor spelling mistake in readme file

uxpanel:
At the top of the readme file it says "x86 Linus" instead of "x86 Linux".


markpizz:
Thank you for identifying this one byte spelling error.  

As it turns out, fixing that is irrelevant at this time since just about to be published changes to this file remove the whole sentence.

PR #1249, opened 2026-05-01
Fixed eclipse.txt file in the NOVA folder.

AyberkTural:
I have changed the NOVA folder files for running properly and "eclipse.txt" file.


markpizz:
I'm trying to understand the value of this set of changes to some 15 files and introducing hundreds of changes across those files.

It looks like you might have started with a zip file at some point, but since you've put together a PR, you must have git running on this system.

On your system, what happens if you:

    $ cd
    $ git clone https://github.com/simh/simh simh-simh
    $ cd simh-simh
    $ make nova
    $# Then run the result:
    $ BIN/nova

What is functionally different from this result compared to what your PR produces?

PR #1251, opened 2026-05-17
Altair8800: Codebase cleanup

deltecent:
fix: eliminate implicit conversions
fix: eliminate C++ keyword usage


deltecent:
Removed AltairZ80 fixes.

PR #1252, opened 2026-06-01
Editing the eclipse.txt file and optimization of switch case

AyberkTural:
PR Title: Refactor: Optimize instruction decoding for the NOVA emulator

Description:
This PR optimizes the instruction decoding logic in the eclipse_cpu.c file.

Changes: Existing if-else structures have been converted to a more efficient and readable switch-case structure.

Purpose: To improve code readability and persistence.

Note: This change alters the core logic of the processor emulator; it only includes structural components of the code.


markpizz:
I ask the same questions as the last time you posted a PR https://github.com/simh/simh/pull/1249:

I'm trying to understand the value of this set of changes to some 15 files and introducing hundreds of changes across those files.

It looks like you might have started with a zip file at some point, but since you've put together a PR, you must have git running on this system.

On your system, what happens if you:

    $ cd
    $ git clone https://github.com/simh/simh simh-simh
    $ cd simh-simh
    $ make nova
    $# Then run the result:
    $ BIN/nova

What is functionally different from this result compared to what your PR produces?

PR #1255, opened 2026-08-09
reverse dependency between pdp11 and pidp11-frontpanel

jdbruner:
A recent commit added a new ``pidp11-frontpanel`` build target. If the build environment is a Raspberry Pi, then when building ``pdp11`` this new build target is also built. There are several concerns with this approach:
1. This makes ``pdp11`` dependent upon ``pidp11-frontpanel``, so that on a Raspberry Pi it is not possible to build ``pdp11`` without also building ``pidp11-frontpanel``. This creates excess unnecessary work (cloning the ``pidp11-frontpanel`` repo and building it even if not needed).
2. There is a breaking change across versions of libgpiod. The API is completely different in libgpiod2 than in libgpiod3. As a result, libgpiod-dev is different across different Debian releases. Specifically, ``pidp11-frontpanel`` uses the newer API and will only build successfully on a Raspberry Pi running Trixie; it will fail on Bookworm (or older) releases, which are still in widespread use. This means that on an older Raspberry Pi, building ``pdp11`` will fail because it depends upon ``pidp11-frontpanel`` -- and that will fail to build.
3. ``GPIO_USEFUL`` is true if the build target is ``pdp11``, but the ``pdp11`` build target doesn't depend upon GPIO; instead, ``pidp11-frontpanel`` does.

The proposed solution is to reverse the dependency so that ``pidp11-frontpanel`` depends upon ``pdp11`` (rather than the other way around). GPIO support is conditioned on the ``pidp11-frontpanel`` build target, not upon ``pdp11``. Add ``pidp11-frontpanel`` to the ``all`` target if and only if ``GPIO_AVAILABLE`` is true.

A fix to ``pidp11-frontpanel`` for the libgpiod breaking change (so that it builds more widely on Raspberry Pi) is outside of the scope of this repository.

### DIFF
--- a/makefile
+++ b/makefile
@@ -250,8 +250,8 @@ endif
 ifneq (,$(or $(findstring pdp8,${MAKECMDGOALS}),$(findstring sds,${MAKECMDGOALS})))
   VIDEO_USEFUL = true
 endif
-# building the PDP11 on certain platforms (Raspberry Pi) could use gpio support
-ifneq (,$(findstring pdp11,${MAKECMDGOALS}))
+# building the PiDP11 front panel on certain platforms (Raspberry Pi) could use gpio support
+ifneq (,$(findstring pidp11-frontpanel,${MAKECMDGOALS}))
   GPIO_USEFUL = true
 endif
 # building the pdp11, any pdp10, any 3b2, or any vax simulator could use networking support
@@ -2643,7 +2643,7 @@ ALL = pdp1 pdp4 pdp7 pdp8 pdp9 pdp15 pdp11 pdp10 \
 	swtp6800mp-a swtp6800mp-a2 tx-0 ssem b5500 intel-mds \
 	scelbi 3b2 3b2-700 i701 i704 i7010 i7070 i7080 i7090 \
 	sigma uc15 pdp10-ka pdp10-ki pdp10-kl pdp10-ks pdp6 i650 \
-	imlac linc tt2500 sel32
+	imlac linc tt2500 sel32 $(if $(GPIO_AVAILABLE),pidp11-frontpanel)
 
 all : ${ALL}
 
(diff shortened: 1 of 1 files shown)


markpizz:
Your fundamental goal is to avoid a failing build, right?

My goal is to successfully build the possibly relevant components if they will build successfully.

When building after a clean clone, the build of the pidp11_frontpanel took about 7 seconds.  Hardly a real concern, right?  
When building in that same directory after a git pull, only the pdp11 simulator will be built with the needed changed pdp11 modules, no mention of pidp11_frontpanel.

The logic in the makefile very specifically tries to only build the pidp11_frontpanel when running on Trixie (aka 13.2).

You say:

> There is a breaking change across versions of libgpiod. The API is completely different in libgpiod2 than in libgpiod3. As a result, libgpiod-dev is different across different Debian releases. Specifically, pidp11-frontpanel uses the newer API and will only build successfully on a Raspberry Pi running Trixie; it will fail on Bookworm (or older) releases, which are still in widespread use. This means that on an older Raspberry Pi, building pdp11 will fail because it depends upon pidp11-frontpanel -- and that will fail to build.
> A fix to pidp11-frontpanel for the libgpiod breaking change (so that it builds more widely on Raspberry Pi) is outside of the scope of this repository.

Please identify what platform you get any build errors on.

I just created a MicroSD with Bookworm on it and cloned the repo and built the pdp11.  There were no build errors on my system.

jdbruner:
I didn't have a spare Pi that I could re-image, so I created a Bookworm container. That effectively resulted in a Bookworm user environment on top of a Trixie kernel. The actual build dependency is upon libgpiod3, so the build failed. I admit this is a corner case; I should have tried a full Bookworm image.

Although I still think the dependency is backwards - the front panel depends upon the PDP11, not the other way around - I'm withdrawing the PR.

PR #1256, opened 2026-08-20
AltairZ80: add the CompuPro CPU-68K machine, CP/M-68K boots from the DISK 1A boot ROM

ajfa:
## What this does

`SET CPU COMPUPRO` gives the M68K CPU a CompuPro CPU-68K board, so AltairZ80
boots CompuPro CP/M-68K off the DISK 1A controller, from the real boot ROM
that is already in the tree.

```
CP/M-68K(tm) Version 1.1
Copyright (c) 1982 Digital Research, Inc.

CompuPro CP/M-68K version 2.1H
Copyright (c) 1983 CompuPro

A>dir
A: CPM      SYS : PIP      68K : STAT     68K : ED       68K : COPY     68K
A: FORMAT   68K : INIT     68K : MFORM    68K : PUTBOOT  68K : DISK2    68K
A: DISK3    68K
A>stat

A: RW, FREE SPACE:     1,026K
```

Configuration:

```
set cpu m68k
set cpu compupro
set ss1 enabled
set mdriveh enabled
set disk1a enabled
set disk1a0 rom
attach disk1a0 cpm68k-a.imd
deposit disk1a bootstrap 8
boot disk1a
```

## Why so little code was needed

Nearly all of it was already here. The 68000 already reaches the S-100 I/O
space through `in()`/`out()`, and `mmiobase`/`mmiosize` already default to the
`FF0000` page that the CPU 68K Technical Manual (A196) gives for memory mapped
I/O. `GetByteDMA`/`PutByteDMA` already send S-100 DMA into 68K memory, so the
floppy controller transfers straight into the 68000's address space. And every
board the CompuPro CBIOS talks to is modelled at the port it expects: DISK 1A
at `C0`, System Support 1 at `50`, DISK 2 at `C8`, DISK 3 at `90`, M-Drive/H
at `C6`, Selector Channel at `F0`, Interfacer 3 at `10`.

What was missing is the boot ROM. A CPU-68K has none of its own in a normal
system. The manual puts it plainly: "In systems with a DISK 1, the DISK 1's
boot ROM will usually be used to provide this data", so the 68000 fetches its
initial stack pointer and program counter from the DISK 1A ROM, which the
board forces onto the bus at address zero until the bootstrap turns it off by
writing to the motor control port.

Bootstrap 8 of that ROM, disassembled, is the whole story:

```
0000  00 00 80 00           initial SSP
0004  00 00 00 0C           initial PC
0008  00 FF 00 00           the I/O base, as data
000C  movea.l $8.l, a3      a3 = 00FF0000
0012  move.l  #$7f, d0
0018  lea.l   $0.l, a1
001E  movea.l #$8000, a2
0024  move.l  (a1)+, (a2)+  copy the ROM to RAM at 8000
0026  dbra    d0, $24
002A  movea.l #$8036, a2
0030  clr.b   $c3(a3)       turn the boot ROM off
0034  jmp     (a2)          continue from RAM
```

## Implementation notes

Three files:

- `s100_disk1a.c` exports the boot ROM so another CPU can read it, and pulses
  a 68000 reset on `BOOT DISK1A` instead of forcing the program counter to
  zero.
- `m68ksim.c` gets the CompuPro mode. The DISK 1A boot ROM answers reads at
  address zero while it is visible, the whole memory mapped I/O window reaches
  `in()`/`out()` for byte, word and long accesses, and the built in MC6850
  console and simulated disk of the generic CP/M-68K machine are switched off,
  since a CompuPro system has its console on a System Support 1 and its disks
  on the S-100 controllers.
- `altairz80_cpu.c` adds `SET CPU COMPUPRO` and `SET CPU NOCOMPUPRO`, and
  reports which machine is selected in `SHOW CPU`.

With `NOCOMPUPRO`, the default, every path is byte for byte what it was, so
the generic CP/M-68K machine is untouched.

## Testing

CompuPro CP/M-68K 2.1H on 8 inch DSDD media, built from the CompuPro CBIOS
sources and the shipped `CPMLDR.SYS` and `CPM.SYS`:

- cold boot from bootstrap 8, five times, always to the `A>` prompt
- `DIR` and `STAT` read the disk, and `STAT` reports the 1,026K the disk
  parameter block says it should
- `PIP` copies between two floppies and the M-Drive/H, and a file written to a
  floppy survives a restart, so the write path reaches the image
- `MFORM` formats the M-Drive/H and programs run from `M:`
- `AS68` assembles the CBIOS sources on drive B: inside the emulated machine,
  and `SIZE` reads the object file back

## AI assistance

Written with Claude Opus 5 (model claude-opus-5, Anthropic) through Claude Code.

### DIFF
--- a/AltairZ80/altairz80_cpu.c
+++ b/AltairZ80/altairz80_cpu.c
@@ -154,6 +154,8 @@ extern t_stat sim_instr_8086(void);
 extern void cpu8086reset(void);
 
 extern unsigned int m68k_cpu_read_byte_raw(unsigned int address);
+extern uint32 m68k_compupro;
+extern t_stat m68k_set_compupro(UNIT *uptr, int32 value, CONST char *cptr, void *desc);
 void m68k_cpu_write_byte_raw(unsigned int address, unsigned int value);
 
 /* function prototypes */
@@ -524,6 +526,10 @@ static MTAB cpu_mod[] = {
         NULL, NULL, "Chooses 8086 CPU"   },
     { MTAB_XTD | MTAB_VDV,  CHIP_TYPE_M68K,     NULL,           "M68K",         &cpu_set_chiptype,
         NULL, NULL, "Chooses M68K CPU"                  },
+    { MTAB_XTD | MTAB_VDV,  1,                  NULL,           "COMPUPRO",     &m68k_set_compupro,
+        NULL, NULL, "Chooses the CompuPro CPU-68K machine for the M68K CPU"     },
+    { MTAB_XTD | MTAB_VDV,  0,                  NULL,           "NOCOMPUPRO",   &m68k_set_compupro,
+        NULL, NULL, "Chooses the generic CP/M-68K machine for the M68K CPU"     },
     { UNIT_CPU_OPSTOP,      UNIT_CPU_OPSTOP,    "ITRAP",        "ITRAP",        NULL, &chip_show,
         NULL, "Stop on illegal instruction"             },
     { UNIT_CPU_OPSTOP,      0,                  "NOITRAP",      "NOITRAP",      NULL, &chip_show,
@@ -6836,6 +6842,7 @@ static t_stat cpu_show(FILE *st, UNIT *uptr, int32 val, CONST void *desc) {
     }
     if (chiptype == CHIP_TYPE_M68K) {
         fprintf(st, "\n\tMemory-Mapped I/O: 0x%06x-0x%06x", mmiobase, mmiobase + (mmiosize - 1));
+        fprintf(st, "\n\tMachine: %s", m68k_compupro ? "CompuPro CPU-68K" : "generic CP/M-68K");
     }
 
     return SCPE_OK;
--- a/AltairZ80/m68ksim.c
+++ b/AltairZ80/m68ksim.c
@@ -160,6 +160,57 @@ extern uint32 in(const uint32 Port);
 
 static uint32 m68k_fc;                              /* Current function code from CPU */
 
+/* CompuPro CPU-68K support.
+ *
+ * The CPU-68K maps the whole S-100 I/O space into extended page FF0000 of
+ * data space (CPU 68K Technical Manual A196, "I/O") and has no console or
+ * disk of its own: everything lives on S-100 boards which AltairZ80 already
+ * models (DISK1A, SS1, DISK2, DISK3, MDRIVEH, SELCHAN, IF3).  In this mode
+ * the built in MC6850 console and the simulated CP/M-68K disk of the generic
+ * machine are switched off so that the entire window reaches in()/out().
+ */
+uint32 m68k_compupro = FALSE;
+
+extern uint32 mmiobase;
+extern uint32 mmiosize;
+
+extern uint32 disk1a_rom_is_visible(void);
+extern uint32 disk1a_rom_base(void);
+extern uint32 disk1a_rom_size(void);
+extern uint32 disk1a_rom_read(uint32 offset);
+
+t_stat m68k_set_compupro(UNIT *uptr, int32 value, CONST char *cptr, void *desc) {
+    m68k_compupro = (value != 0);
+    return SCPE_OK;
+}
+
+/* Returns the byte the DISK 1A boot ROM answers with, or -1 when the ROM is
+   not visible at this address. */
+static int32 m68k_compupro_rom(uint32 address) {
+    uint32 base;
+    if (!m68k_compupro || !disk1a_rom_is_visible())
+        return -1;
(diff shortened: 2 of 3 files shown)


markpizz:
Hi there @ajfa,

These changes look reasonable to me, but @psco approves updates to this simulator.

Meanwhile, your identity AJFA, is less than useful since, in general we like to be able to identify who's made and/or proposed changes to the codebase.  Please put your whole name in your github profile and preferably also use your email address in the commits you are proposing.

The same commentary applies for PR #1257.

Thanks,

- Mark

psco:
Without the referenced disk images I cannot say anything about the quality of this PR. I am also a bit hesitant to waste my time on bot generated stuff.

ajfa:
@psco On the images: `cpm68k-a.imd` is not a preserved dump, which is why there is nothing to find. It is generated from material that is already public: the CompuPro software archive for the files, Dave Dunfield's CompuPro masters to establish the boot track layout, and the DISK 1A boot ROM, which needs no supplying at all because it is already in this tree as `disk1a_rom[16][512]` in `AltairZ80/s100_disk1a.c`.

Rather than ask you to take my word for the image, here is the builder that produces it: **https://github.com/ajfa/compupro-simh**. Two files of Python with nothing outside the standard library, plus a script per family that takes the path to the unzipped archive and writes the six images these two PRs use. I am handing over the recipe rather than the image because the disk carries Digital Research and CompuPro software and I would rather not redistribute that. If you would still prefer the image itself, say so and I will attach it here.

The geometry, in case it is quicker to check than to run: 8 inch, 77 cylinders, two heads. The data area is CompuPro density 3, MFM with 8 sectors of 1024 bytes. Cylinder 0 is FM with 26 sectors of 128 bytes, on both heads for the CP/M-68K layout, which is what the CP/M-68K `FORMAT` program writes, and on head 0 only for the CP/M-86 one. CP/M sees one logical track per (cylinder, head) pair, head = track & 1 and cylinder = track >> 1. The data area is 2048 byte blocks, 600 blocks, 256 directory entries and 4 reserved tracks, with the skew the CBIOS uses.

On the other point, the PR body already says under "AI assistance" how this was written, because I would rather you knew than guessed. In place of trust, here is what was actually exercised: cold boot from bootstrap 8 five times to the `A>` prompt; `STAT` reporting the 1,026K the disk parameter block says it should; `PIP` copying between two floppies and the M-Drive/H, with the file surviving a restart; `MFORM` formatting the M-Drive/H; and `AS68` assembling the CBIOS sources on drive B: inside the machine, with `SIZE` reading the object file back. If any of that does not reproduce for you, that is a defect and I want to hear about it.

@markpizz On identity: `ajfa` is a stable GitHub account rather than a throwaway, and the commits are verifiable against it. I keep my email address private as a matter of course and would rather not change that. If a name on the profile is what matters for attribution I can look at that, but I would ask that the account itself be enough for the code review. The same goes for #1257.

And if that rule means these changes cannot be taken on, I understand, and there are no hard feelings. The reasoning and the measurements are in the PR either way, so the work is there for whoever wants to redo it under their own name.

psco:
The image builder is quite useful, thank you @ajfa.

Unfortunately this PR does not pass my personal quality criteria:

1. Does the PR fix a real, reproducible defect? The answer is no.
2. Or, does the PR allow software to run that was not possible before? Again the answer is no. Using the configuration:
```
set cpu m68k
set disk1a enabled
att disk1a0 cpm68k-a.imd
set ss1 enabled
set mdriveh enabled
load CPRO68K.BIN 0
go c
```
with CPRO68K.BIN from my site you get
```
CP/M-68K(tm) Version 1.1 
Copyright (c) 1982 Digital Research, Inc.

CompuPro CP/M-68K version 2.1H
Copyright (c) 1983 CompuPro

A>MFORM M X

A>dir
A: CPM      SYS : PIP      68K : STAT     68K : ED       68K : COPY     68K 
A: FORMAT   68K : INIT     68K : MFORM    68K : PUTBOOT  68K : DISK2    68K 
A: DISK3    68K 
A>stat

A: RW, FREE SPACE:     1,026K
```
without any changes to AltairZ80.
3. Do the design  and the code appear to be sound? The code looks somehow ok but I would not call adding almost 150 lines of code "little". But I am not convinced that adding the "compupro" switch and modifying the 68k simulator code was the right design decision.
4. Has the code been properly (regression) tested? I doubt that proper testing has taken place, in particular regression testing appears to be missing. In the case of bot assisted (or written) code resp. PR, each and every statement made must be verifiable as the concept of trust has, at least for me, vanished.
5. What is the reputation of the contributor? Since this seems to be a first time contribution to AltairZ80 this is hard to judge. But a real name would help.

In summary: This PR should **not** be merged.

PR #1259, opened 2026-08-23
Add PDP-11 front panel client and RT-11/BASIC-11 demo setup

rolandkirsche:
## Summary
- Add `frontpanel/PDP11Panel.c`, a front panel client driving a local PDP-11/70 RT-11 setup via sim_frontpanel (PC/PSW/SP/R0-R5 live display, BOOT/CONT/STEP/BREAK/EXAMINE)
- Fix `OS_CURSES_DEFS` in the makefile bundling a link-only `-lncurses` flag into compile-step OPTS, which errors under `-Werror` on macOS; split into `OS_CURSES_DEFS` (compile) and `OS_CURSES_LDFLAGS` (link)
- Drop the ncurses rendering path in the panel client (it garbled output because printf wasn't redirected through curses) in favor of the working plain ANSI CSI escape path
- Add a working RT-11 V4 + BASIC-11 V2.1 demo setup: `configs/rt11.ini`, `configs/rt11-basic.ini`, `disks/rt11-v4.dsk`, `disks/basic-11.dsk`, `start.sh`

## Test plan
- [x] Built via `make pdp11 pdp11panel` on macOS (Darwin), confirmed no more `-Werror`/link errors from the curses flag split
- [x] Booted `./start.sh` (`configs/rt11.ini`), confirmed RT-11SJ V04.00C boots and `DIR` lists the RK0 volume correctly
- [x] Booted `configs/rt11-basic.ini`, confirmed RK1 (BASIC-11 image, read-only) is attached and `RUN RK1:BASIC` starts BASIC-11/RT-11 V2.1 and executes a test program correctly
- [x] Ran `pdp11panel` against the running simulator and confirmed live register display without the garbled-curses issue

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01EVsHDP6P8boNt3fS1gNxpf


markpizz:
Hi Roland,

Thank you for finding & fixing the make bug relating to the frontpaneltest program.  The sim_frontpanel API (and thus the need to run the test program) hasn't changed since before the makefile was enhance to support building modules separately.  Your fix has been merged into the makefile with you as the author.

Meanwhile, the use of the sim_frontpanel API was designed specifically so folks could implement either frontpanel GUI or physical front panel interfaces for simulators and still run unmodified simulators to achieve this.

The frontpaneltest program exists very specifically to test and debug the sim_frontpanel API and it only used the vax simulator behind the scenes since that simulator has a built-in ROM which runs when the CPU is booted and thus a running simulator is available to test and exercise the API without  depending on any external support files.

The pdp11panel application you've implemented doesn't look or behave like anything which existed on any PDP11 system, but if you want to go down that path, that's fine.

I've taken your pieces (with one fix in PDP11Panel.c removing an unused variable i) and organized those pieces the way you had, but in a separate repository which uses the simh/simh repo as a submodule.

You should be able to pickup this repo with

    $ git clone --recurse-submodules https://github.com/simh/simh-panels
    $ cd simh-panels
    $ gmake pdp11panel {or any other normal simh target}

Some bug fixes to the simh_frontpanel API and the frontpaneltest are coming shortly.

Thoughts?

- Mark

markpizz:
Hello again Roland,

I reorganized the simh-panels repo a little crediting you for PDP11Panel.

Thoughts?

